PR #34, opened 2021-10-02
Update gradle version to 7.0.2

AnkitSuda:
Update kotlin version to 1.5.31


LagradOst:
No, this wont be accepted because it does not allow me to work in intelij and that is my main editor for android.

PR #35, opened 2021-10-02
Design fixes

AnkitSuda:
Changed search results card spacing
Changed tag spacing


AnkitSuda:
I can revert version upgrades if you want.

LagradOst:
I dont get what this pr is trying to do. It removed the margins on advanced search and added margins on tags, making both look worse. (Also 8.toDp converts px to dp when I think you want to convert to Px because most android code uses dp in xml but px in code or that is my guess)

PR #84, opened 2022-05-19
New Provider Ranobes

KillerDogeEmpire:


KillerDogeEmpire:
This adds a new provider. If you have any changes please let me know

KillerDogeEmpire:
It works fine. The only book that triggers anti bot I found is Supreme Magus. Tomorrow I will fix that

app/src/main/java/com/lagradost/quicknovel/providers/RanobesProvider.kt
```diff
@@ -0,0 +1,264 @@
...
+        Pair("Yaoi", "Yaoi"),
+    )
+
+    override fun loadMainPage(
+        page: Int,
+        mainCategory: String?,
+        orderBy: String?,
+        tag: String?
+    ): HeadMainPageResponse {
+        val url = "$mainUrl/tags/genre/$tag/"
```
LagradOst: This was the bug in the previus providers you merged, you have to take into account the page number by adding "page/$page/"

app/src/main/java/com/lagradost/quicknovel/providers/RanobesProvider.kt
```diff
@@ -0,0 +1,264 @@
...
+                )
+            )
+        }
+        return HeadMainPageResponse(url, returnValue)
+    }
+
+    override fun loadHtml(url: String): String? {
+        val response = khttp.get(url)
+        val document = Jsoup.parse(response.text)
+        Thread.sleep(Random.nextLong(45,90))
```
LagradOst: You can override the delay between requests rateLimitTime instead of doing it here

LagradOst (changes requested):

app/src/main/java/com/lagradost/quicknovel/providers/RanobesProvider.kt
ghost: I think that it has to stay like that. I will ask Addipe about that. I did make it able to load novels with 1000+ chapters now

PR #103, opened 2022-07-19
New Provider requested by 𝑨𝒌𝒊𝒉𝒊𝒓𝒐#7730

KillerDogeEmpire:


KillerDogeEmpire:
This is ready to merge, just let me know if there are any changes needed

app/src/main/java/com/lagradost/quicknovel/providers/AllNovelProvider.kt
```diff
@@ -80,6 +80,7 @@ class AllNovelProvider : MainAPI() {
         orderBy: String?,
         tag: String?
     ): HeadMainPageResponse {
+        val url = "$mainUrl/$tag/page/$page/"
```
LagradOst: url gets redefined, later. so this line does nothing

app/src/main/java/com/lagradost/quicknovel/providers/NovelFullProvider.kt
```diff
@@ -0,0 +1,236 @@
...
+        Pair("Canopy", "Canopy")
+    )
+
+    override fun loadMainPage(
+        page: Int,
+        mainCategory: String?,
+        orderBy: String?,
+        tag: String?
+    ): HeadMainPageResponse {
+        val url = "$mainUrl/$tag/page/$page/"
```
LagradOst: url gets defined later so this does nothing, please check the site how they do this

ghost: @LagradOst I did it the same way I did with AllNovel

app/src/main/java/com/lagradost/quicknovel/providers/NovelFullProvider.kt
```diff
@@ -0,0 +1,235 @@
...
+            "Ongoing" -> STATUS_ONGOING
+            "Completed" -> STATUS_COMPLETE
+            else -> STATUS_NULL
+        }
+
+        var rating = 0
+        var peopleVoted = 0
+        try {
+            rating =
+                document.selectFirst(" div.small > em > strong:nth-child(1) > span").text().toInt()
```
Blatzar, line 214: You can do 
```kotlin
val rating = document.selectFirst(" div.small > em > strong:nth-child(1) > span").text().toIntOrNull() ?: 0
```
To avoid exceptions

Blatzar: No need to edit, just letting you know cool features

app/src/main/java/com/lagradost/quicknovel/providers/NovelFullProvider.kt
```diff
@@ -0,0 +1,235 @@
...
+        val parsed = parsedchaptersData.select("select > option")
+        for (c in parsed) {
+
+            val cUrl = mainUrl + c.attr("value")
+            val cName = if (c.text().isEmpty()) {
+                "chapter $c"
+            } else {
+                c.text()
+            }
+            data.add(ChapterData(cName, cUrl, null, null))
```
Blatzar, line 199: instead of adding to an array it's cleaner to use maps most time, like:
```kotlin
 val data = parsed.map { c ->
            val cUrl = mainUrl + c.attr("value")
            val cName = if (c.text().isEmpty()) {
                "chapter $c"
            } else {
                c.text()
            }
            ChapterData(cName, cUrl, null, null)
}
```
No need to change tho

Blatzar: (This can be done pretty much anywhere where you iterate over items and add them to a list)

app/src/main/java/com/lagradost/quicknovel/providers/NovelFullProvider.kt
```diff
@@ -0,0 +1,235 @@
...
+        mainCategory: String?,
+        orderBy: String?,
+        tag: String?
+    ): HeadMainPageResponse {
+        val firstresponse = khttp.get(mainUrl)
+        val firstdocument = Jsoup.parse(firstresponse.text)
+        fun getId(tagvalue: String?): String? {
+            for (i in firstdocument.select("#hot-genre-select>option")) {
+                if (i.text() == tagvalue) {
+                    return i.attr("value")
```
Blatzar, line 88: Instead of looping like this you can use .first

I think you can do something like this:
```kotlin
return firstdocument.select("#hot-genre-select>option").firstOrNull { it.text() == tagvalue }?.attr("value")
```

PR #105, opened 2022-07-23
New Providers

KillerDogeEmpire:


LagradOst (approved):
Very good, ready to merge? 👍

The icons are the same tho, is that intended?

KillerDogeEmpire:
Yes, it's ready to merge. The icons are the same because they have the same icon on the website

PR #116, opened 2022-08-26
Fix freewebnovel

KillerDogeEmpire:
Before
unknown1
unknown

After 
Screenshot 2022-08-25 180553


AbdullahM0hamed:
Lowkey thought he was gonna say use regex

PR #148, opened 2023-01-11
Improved the epub* generation and fixed poster refreshing

Blatzar:
This also includes some fixes for the adapter


LagradOst:
very cool

PR #209, opened 2023-12-01
[New Provider] `libread.com`

USLTD:
I have added `libread.com` which is almost the same as `freewebnovel.com` with few changes.
So, essentially this provider is a copy-paste of [**FreeWebNovelProvider.kt**](https://github.com/LagradOst/QuickNovel/blob/master/app/src/main/java/com/lagradost/quicknovel/providers/FreeWebNovelProvider.kt), but with few basic fixes.

Currently I haven't implemented rating fetching, so it always stays 0, but it doesn't require much to implement it, so either I (when I will have free time again), or someone else will implement it.

Additionally, I have only tested this using my physical device (I couldn't get the emulator start no matter what I tried to do so), so I can't guarantee that it will work on every device.


LagradOst:
Providers dont use device specific functions/api levels, so dont worry about not using an emulator as the core provider code will work on every device

USLTD:
> Providers dont use device specific functions/api levels, so dont worry about not using an emulator as the core provider code will work on every device

I know but still. Oh, and just in case. I have compiled my version using the android-34 platform (although I set the target to 33), not lower since I didn't have enough time to install another platform, but I don't think it is enough to change something drastically, right? 😥

LagradOst (approved):
Code works, however it fucks up with tags as scrolling will return the exact same results. return EmptyList or make genres scrollabe

USLTD:
Sorry @LagradOst, but for some reason my **Android Studio** got fucked and throws dumb errors in editor. So, before I manage to fix this, I won't be able to test any code.

USLTD:
PR is ready.

LagradOst (approved):
Works 👍 
If you want this merged then reactor with or comment

USLTD:
MERGE!!!!

PR #217, opened 2024-01-21
Remove warnings (this belongs on rr, report on amazon etc) from RR

AbdullahM0hamed:
"blah blah blah RoyalRoad blah blah blah Amazon blah blah blah report"

"Sir this is a cultivation novel"

Anyways man, didn't know you were Jeff Bezos

### DIFF
--- a/app/src/main/java/com/lagradost/quicknovel/providers/RoyalRoadProvider.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/providers/RoyalRoadProvider.kt
@@ -375,6 +375,8 @@ class RoyalRoadProvider : MainAPI() {
     override suspend fun loadHtml(url: String): String? {
         val response = app.get(url)
         val document = Jsoup.parse(response.text)
-        return document.selectFirst("div.chapter-content")?.html()
+        val chap = document.selectFirst("div.chapter-content")
+        chap?.select("p:not([style])")?.remove()
+        return chap?.html()
     }
 }


AbdullahM0hamed:
I made an assumption that the pattern I noticed would be consistent and boy was I wrong

app/src/main/java/com/lagradost/quicknovel/providers/RoyalRoadProvider.kt
```diff
@@ -375,6 +375,18 @@ class RoyalRoadProvider : MainAPI() {
     override suspend fun loadHtml(url: String): String? {
         val response = app.get(url)
         val document = Jsoup.parse(response.text)
-        return document.selectFirst("div.chapter-content")?.html()
+        val styles = document.select("style")
+        val hiddenRegex = Regex("^\\s*(\\..*)\\s*\\{")
+        val chap = document.selectFirst("div.chapter-content")
+        styles.forEach { style ->
+            hiddenRegex.findAll(style.toString()).forEach {
```
AbdullahM0hamed, line 382: Ideally, this would be `style.text()`, but that gives an empty string/null and results in the whole thing just doing nothing

app/src/main/java/com/lagradost/quicknovel/providers/RoyalRoadProvider.kt
```diff
@@ -375,6 +375,18 @@ class RoyalRoadProvider : MainAPI() {
     override suspend fun loadHtml(url: String): String? {
         val response = app.get(url)
         val document = Jsoup.parse(response.text)
-        return document.selectFirst("div.chapter-content")?.html()
+        val styles = document.select("style")
+        val hiddenRegex = Regex("^\\s*(\\..*)\\s*\\{")
```
AbdullahM0hamed: Should be a multiline regex that checks `display: none`, but I couldn't get that to work and cba to play around until kotlin let's me do multiline. (I did try RegexOption.MULTILINE though, so it either isn't good enough or morely likely the regex I tested in python just doesn't work here fsr)

AbdullahM0hamed:
Ffs this isn't working, should've kept better track of the changes I made after testing that I then manually changed back (wish I'd kept the apk too)

AbdullahM0hamed:
At least I kept something tangible that shows I didn't hallucinate the whole "it worked" thing (word count difference, didn't think to actually show examples of missing amazon warnings, but I did check at the time)

AbdullahM0hamed:
Once again:

> I made an assumption

And was wrong. I really thought it's all in one line, and the multiple line regex (I believe it was an addition similar to `\\.*display:\\s*none`) didn't work, surely we don't need `RegexOption.MULTILINE` then.

LagradOst:
Holy fuck you made a pr 🥳 
I will merge this after testing it, eta 1-2 days

LagradOst:
I do not know how this works, but it works

PR #259, opened 2024-08-13
New provider ( mtlarchive)

Legendoras:


LagradOst (commented):
Bro you cant just copy paste the royalroad provider and change the url, that is now how this works at all.

PR #284, opened 2025-01-10
Restore MtlNovelProvider

1aam2am1:
This commit restores MtlNovelProvider, that was commented some time ago.
As some novels are only mtl translated. 
Additionally removed Cloudflare, as site is no longer using it.


LagradOst:
Very nice 👍

Issue #272, opened 2024-11-06
Request: Chapter list QoL features


KrisKamweru:
I reckon it would be nice to be able to reverse the order of the list, filter by read status, and mark all chapters before selected chapter as read.

Issue #73, opened 2022-04-29
[Feature Request] Allow to Bookmark a selected chapter


seizue:
Hi. Can you add a feature to bookmark a selected chapter like in the SS.
Aside from bookmark feature, Filter Bookmark would be nice addition.. ^^

![IMG_20220429_194855](https://user-images.githubusercontent.com/25120376/165939519-0193a33f-3345-4c55-a4e0-5e09a935878a.jpg)

![IMG_20220429_194955](https://user-images.githubusercontent.com/25120376/165939583-5380e49d-764b-4242-9c8c-521e0c62f4fe.jpg)

![IMG_20220429_195748](https://user-images.githubusercontent.com/25120376/165939932-e6e37c05-fe13-4815-abbf-32553c346c31.jpg)

PR #334, opened 2025-07-08
Add context menu for chapter selection and bookmark functionality/tab

AtaraxyState:
This was originaly made for my personal use only, but I figured you might want to add it to main app. Tell me if anything is wrong here or if I should change smt.

### DIFF
--- a/QuickNovel.code-workspace
+++ b/QuickNovel.code-workspace
@@ -1,8 +1,0 @@
-{
-	"folders": [
-		{
-			"path": "."
-		}
-	],
-	"settings": {}
-}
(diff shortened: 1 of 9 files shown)


AtaraxyState:
I'll try to find issues linked to this.
https://github.com/LagradOst/QuickNovel/issues/73
https://github.com/LagradOst/QuickNovel/issues/272 (partly)

AtaraxyState:
Maybe instead of a bookmark tab we could add a filter option in chapter tab, like in most other apps. I wanted smt easy to implement for a start but I could work on it if you want.

LagradOst:
While this PR is a good addition, it could be improved. For example, right now you have a separate tab, even if it displays more or less the same information, just filtered. So a filter+sort on the current tab like #73 makes more sense. This way you could also implement so it is more data driven, more or less just modifying the observed chapters like right now, instead of "triggerChapterRefresh".

PR #373, opened 2025-11-21
Added RR author notes

Azykle:
Author notes sometimes contain crucial context. For instance, an author might only reference significant changes in the previous chapter within the author notes.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -22,7 +22,7 @@ android {
         minSdkVersion 21
         targetSdkVersion 35
         versionCode 56
-        versionName "3.3.9"
+        versionName "3.4.0"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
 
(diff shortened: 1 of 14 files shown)


LagradOst:
I know that the current build system is broken, however I would much rather that you created a separate pull request for it. Moreover, the "pre-release" apk work is not signed. If you where to create a pre-release workfile, then take a look at https://github.com/recloudstream/cloudstream/blob/master/.github/workflows/prerelease.yml as key-signing is not something you can add after someone has downloaded the first version.

LagradOst (commented):
This is very cool, but it would be every better if the user could customize this. So I want you to make it so I can filter out or query the notes if I was to add it to a future update.

To make it as futureproof and easy to work with as possible I would like you to format the spans somewhat like this:
```
<div class="qnauthorsnoteheader">

<div class="qnauthorsnote">NOTE HERE</div>
<div class="qnauthorsseperator"><p>━━━━━━━━━━━━━━━━━━━━</p><p>&nbsp;</p></div>
...
</div>
```

It should not affect the epub reading experience, but allows me to change the format later on already generated epubs or downloaded content.

LagradOst (approved):
This looks good, I will test this and possibly even add a setting to enable/disable this in the reader too.

PR #395, opened 2026-01-09
I added the optional online translation, which is better than offline

TJMolina:
I added the optional online translation, which is better than the offline translation and has no request limit (or it's very large, based on my testing). I also reduced the chapter preloading to 1 in both directions, previous and next.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -64,7 +64,7 @@
     <string name="remove_bloat">Remove Bloat</string>
     <string name="remove_bloat_desc">Removes the title and translator from all pages, will cause longer generation time</string>
     <string name="check_for_update">Check for Update</string>
-    <string name="translating">Translating chapter</string>
+    <string name="translating">Translating</string>
 
 
 


app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -878,51 +890,83 @@ class ReadActivityViewModel : ViewModel() {
...
+                } else {
+                    emptyList()
+                }
+
+                for (j in batch.indices) {
+                    val originalText = batch[j].text.toString()
+
+                    // Fallback: If translation fails or a paragraph is missing, try offline translation or use the original text
+                    // This block now handles both: Offline mode by default OR Online mode failure
+                    val finalText = translatedParagraphs.getOrNull(j) ?: try {
```
LagradOst: I do not think it should use the fallback at all. As it is invisible to the user, and gets included into the cache even if it is wrong. Showing an error message is better for the user to indicate that we lack network or something.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -878,51 +890,83 @@ class ReadActivityViewModel : ViewModel() {
...
-                builder.append('\n')
-                out.add(TextSpan(newText.toSpanned(), start, end, span.index, span.innerIndex))
+            val separator = "\n\n" // Use double line breaks to separate paragraphs within the batch
+            val onlineTranslationIsOn = currentSettings.useOnlineTranslation
+
+            // Use batchSize 5 for online, 1 for offline to unify the loop logic
+            val batchSize = if (onlineTranslationIsOn) 5 else 1
+
+            for (i in 0 until spans.size step batchSize) {
+                loading.invoke(Triple(spans[i].index, i, spans.size))
```
LagradOst: i is not necessarily innerIndex

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -878,51 +890,83 @@ class ReadActivityViewModel : ViewModel() {
...
-                val start = builder.length
-                builder.append(newText)
-                val end = builder.length
-                builder.append('\n')
-                out.add(TextSpan(newText.toSpanned(), start, end, span.index, span.innerIndex))
+            val separator = "\n\n" // Use double line breaks to separate paragraphs within the batch
+            val onlineTranslationIsOn = currentSettings.useOnlineTranslation
+
+            // Use batchSize 5 for online, 1 for offline to unify the loop logic
+            val batchSize = if (onlineTranslationIsOn) 5 else 1
```
LagradOst: Just write two different loops, as the loop is 10 lines long.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -878,51 +890,83 @@ class ReadActivityViewModel : ViewModel() {
...
-                    Log.i(TAG, "Cached $filePrefix")
+                    Log.i(TAG, "Successfully cached $filePrefix")
                 }
             }
 
             return mlRawText.toSpanned() to out
-        } catch (t: ExecutionException) {
-            throw t.cause ?: t
+        } catch (t: Exception) {
+            logError(t)
```
LagradOst: No, we do not want to catch everything like this and rethrow. The reason why we rethrow the ExecutionException is to make the error message better.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -960,23 +1004,27 @@ class ReadActivityViewModel : ViewModel() {
...
-            val keys =
-                chapterData.keys.toTypedArray() // deep copy it to avoid ConcurrentModificationException
+            val keys = chapterData.keys.toTypedArray() // deep copy it to avoid ConcurrentModificationException
 
             // remove all irrelevant cache so we do not translate outdated shit
             for (key in keys) {
-                if (key < lower || key > upper) {
-                    chapterData.remove(key)
+                //// If online translation is enabled, reduce chapter preloadin
+                if(mlSettings.useOnlineTranslation)
```
LagradOst: This change is unessessary. The old code here should work fine.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -1908,4 +1969,47 @@ class ReadActivityViewModel : ViewModel() {
...
+        val baseUrl = "https://translate.googleapis.com/translate_a/single"
+
+        return try {
+                if (text.trim().isBlank()) return null
+
+                val response = MainActivity.app.get(
+                    "$baseUrl?client=gtx&sl=auto&tl=$targetLang&dt=t&q=${Uri.encode(text)}"
+                ).text
+
+                val jsonArray = JSONArray(response)
```
LagradOst: We have a json parse extention like `app.get(...).parsed<T>()` where T is the json class you get. It would make the code a lot cleaner!

.idea/markdown.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
LagradOst: Please remove all .idea files

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -514,7 +520,7 @@ class ReadActivityViewModel : ViewModel() {
     private var chapterPaddingBottom: Int = 1
 
     /** upper padding, for preloading current+chapterPaddingTop */
-    private var chapterPaddingTop: Int = 1
+    private var chapterPaddingTop: Int = 2
```
LagradOst, line 526: Why?

LagradOst: Is this not the opposite of what you wrote?

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -884,76 +890,80 @@ class ReadActivityViewModel : ViewModel() {
             context?.cacheDir?.let {
                 val cache = File(it, "$filePrefix.txt")
                 if (cache.exists()) {
-                    Log.i(TAG, "Cache exists for $filePrefix")
```
LagradOst, line 881: Why did you remove this line?

app/src/main/res/layout/read_bottom_settings.xml
```diff
@@ -467,8 +467,20 @@
                         style="@style/WhiteButton"
                         android:layout_weight="1"
                         android:layout_width="wrap_content" />
+
             </LinearLayout>
 
+            <com.google.android.material.checkbox.MaterialCheckBox
```
LagradOst, line 473: `style="@style/WhiteCheckmark"`

app/src/main/java/com/lagradost/quicknovel/MainActivity.kt
```diff
@@ -105,6 +105,11 @@ class MainActivity : AppCompatActivity() {
             mainActivity?.openEpubPicker()
         }
 
+        //+++++++++++++++++++++++++++++++++++++++++++++++++++++//
+        //TODO try to add a function to import PDFs
```
LagradOst, line 109: Wrong PR?

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: Okayyy, I'll make the changes, I'm a bit of a noob sorry haha

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: "I know, i goes in steps of 5. This is to merge paragraphs and make fewer requests to the API, and it also speeds up the translation. I create blocks of 5 further down in val batch = spans.subList(i, minOf(i + batchSize, spans.size)) and then split them back into 5 different paragraphs

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: "Haha, I love shortening code, sorry! But if you prefer, I can split it into two separate loops. Or is that how it should be done in the first place? I'm still a beginner

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: Okk, I'll make the changes

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: I forgot to change that. I had modified it earlier because I couldn't think of a way to shorten the translation time

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: While testing, occasionally a bug occurs where, for example, chapter 3 gets translated before chapter 2 while reading if I set the chapterPaddingTop to 1...

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: My mistake, sorry about that

app/src/main/java/com/lagradost/quicknovel/MainActivity.kt
TJMolina: I don't remember making that change, I thought that comment was already there. Although in fact I am going to try to add the function to import PDFs

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: Or if you mean the chapter index that shows which one is being translated, I already fixed that but haven’t uploaded it yet, I was making more changes

app/src/main/res/values/strings.xml
```diff
@@ -249,4 +273,11 @@
     <string name="import_epub">Import epub</string>
     <string name="show_authors_notes">Show authors notes</string>
 
+    <!-- Some errors -->
```
LagradOst, line 276: Right now these are unused? Additionally, did you mean to place these should be placed in the xml for Spanish. 

Otherwise I think you can remove these,

app/src/main/res/values/strings.xml
```diff
@@ -49,6 +49,26 @@
     <string name="select_chapter">Select Chapter</string>
     <string name="poster_descript">Poster</string>
     <string name="sort_downloads_descript">Sort Downloads</string>
+    <string name="search_roviders">Search Providers</string>
+
+    <string name="search">Search</string>
+    <string name="general">General</string>
+    <string name="info">Info</string>
+
+    <string name="advanced_search">Advanced Search</string>
```
LagradOst, line 58: These are unused, did you mean to also change settings.xml?

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -884,45 +899,78 @@ class ReadActivityViewModel : ViewModel() {
...
+            } else {
+                // --- Offline mode ---
+                for (i in spans.indices) {
+                    loading.invoke(Triple(spans[i].index, i, spans.size))
+                    val originalText = spans[i].text.toString()
+
+                    // Si falla el modelo offline, usamos el original
+                    val finalText = try {
+                        Tasks.await(translator.translate(originalText))
+                    } catch (e: Exception) {
```
LagradOst: This should be:
```kotlin
} catch (t: ExecutionException) {
    throw t.cause ?: t
}
```

To show the error correctly.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -884,45 +899,78 @@ class ReadActivityViewModel : ViewModel() {
...
+                for (i in 0 until spans.size step batchSize) {
+                    loading.invoke(Triple(spans[i].index, i, spans.size))
+                    val batch = spans.subList(i, minOf(i + batchSize, spans.size))
+
+                    val combinedText = batch.joinToString(separator) { it.text.toString() }
+                    val translatedBatch = onlineTranslate(combinedText, currentSettings.to)
+                    val translatedParagraphs = translatedBatch?.split(separator) ?: emptyList()
+
+                    for (j in batch.indices) {
+                        val finalText = translatedParagraphs.getOrNull(j) ?: batch[j].text.toString()
```
LagradOst, line 930: I still think this should throw an error instead so it can be shown to the user. Otherwise the user may get confused why they are not translated. So something like `?: throw ErrorLoadingException("Unable to translate")`.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -884,45 +899,78 @@ class ReadActivityViewModel : ViewModel() {
...
+                    val end = builder.length
+                    builder.append('\n')
+                    out.add(TextSpan(finalText.toSpanned(), start, end, spans[i].index, spans[i].innerIndex))
+                }
             }
 
             val mlRawText = builder.toString()
 
-            // atomically write the file
+            // Guardar en caché de forma atómica
```
LagradOst: Why is this translated? Please make the comments in english so everyone can understand it.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
LagradOst: This is still unchanged.

app/src/main/res/values/strings.xml
TJMolina: I will translate it into English and leave it there in values. Afterwards, I will do the same but in values-es

TJMolina: These are errors that are written directly in the code instead of using R.strings

app/src/main/res/values/strings.xml
TJMolina: I’m going to use them too. It’s just that in some XML files they used text directly instead of the string resource, so I created them

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: Thanks, I’ll do

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: It’s difficult for me right now, I’ve been studying the code. The problem is that when the translator throws an error, for example from not having Wi-Fi, if I stop translating the loader stays and the chapter doesn’t load. I’m still slowly figuring out how it’s done. Although I think simply clearing the cache and re-entering the chapter is easier, I just don’t know how easy that would be for the average user…

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: Sorry, I don’t speak English either and sometimes I translate the comments haha. I’ll put it back in English and won’t touch it anymore

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: But I think that non-native English-speaking users will still really enjoy the online mode as it is now haha

TJMolina:
Forget what I said, it wasn’t difficult to change haha. What I did is that if the online translation fails, I clear the cache and show the original text along with a toast. This way, when the user reconnects to the internet, they can try translating online again

TJMolina:
Done

PR #397, opened 2026-01-13
Epub bugs fix and status bar

TJMolina:
### 🔧 Changes Made
- Fixed errors in the import of EPUBs that occur when poorly generated EPUBs are imported.  
- Fixed the blue progress bar for novels so that it shows **reading progress** instead of only **download progress**.  
- Added more functionality to improve the user experience.  

### 📊 Progress Bars
- **White status bar** → Chapters downloaded  
- **Blue status bar** → Chapters read  

### 📸 Screenshots
WhatsApp Image 2026-01-13 at 17 27 53  
WhatsApp Image 2026-01-13 at 17 19 55

### DIFF
--- a/app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
@@ -20,7 +20,8 @@ import androidx.core.text.toSpanned
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
-import androidx.lifecycle.viewModelScope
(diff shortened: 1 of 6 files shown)


LagradOst:
No, the blue progress bar is used for partial downloads. Do not override it with reading.

TJMolina:
I see, so can I add a third bar? Or do you think it's a useless feature?

LagradOst:
> Okay, I’ll delete it, but what do you think about the epub part? The errors happen, for example, when you convert a pdf into an epub or download a low-quality epub and still import it. The adjustments I made allow poorly imported epubs to still have a cover and chapters.

Ye that is a nice change. I have not tried poorly imported epubs so I do not know the details here.

PR #398, opened 2026-01-15
Importar pdf

TJMolina:
I added the option to import PDFs. Due to the complexity of reading them, they are also converted into EPUBs when saved, so the process of reading a PDF—which is significantly more costly than reading an EPUB—doesn’t have to be repeated

### DIFF
--- a/app/src/main/java/com/lagradost/quicknovel/ui/download/AnyAdapter.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/ui/download/AnyAdapter.kt
@@ -257,6 +257,8 @@ class AnyAdapter(
                     val same = imageText.text == card.name
                     backgroundCard.apply {
                         setOnClickListener {
+                            if(card.apiName == IMPORT_SOURCE_PDF && card.downloadedCount < card.downloadedTotal)
+                                preloadPartialImportedPdf(card, context)
                             downloadViewModel.readEpub(card)
                         }
                         setOnLongClickListener {


app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1619,11 +1629,232 @@ object BookDownloader2 {
...
+
+        try {
+            val outStream = ByteArrayOutputStream()
+
+            //get image compressed as byteArray
+            bitmap.compress(Bitmap.CompressFormat.JPEG, 80, outStream)
+            val imgBytes = outStream.toByteArray()
+
+            // only over 90KB images
+            if (imgBytes.size >= 90 * 1024) {
```
LagradOst: Why this arbitrary limit?

TJMolina: The thing about only 90kb images is to avoid annoying icons that are used for headers and footers. I’m attaching examples

In the original pdf:
image


in the app without the limit of 90kb:
image


in the app with the limit:

image

app/src/main/java/com/lagradost/quicknovel/ui/download/DownloadFragment.kt
```diff
@@ -114,7 +115,7 @@ class DownloadFragment : Fragment() {
             return id
         }
 
-        val isImported: Boolean get() = apiName == IMPORT_SOURCE
+        val isImported: String get() = apiName
```
LagradOst: This only reexports the apiName, did you mean to write `apiName == IMPORT_SOURCE || apiName == IMPORTED_PDF`

app/src/main/java/com/lagradost/quicknovel/ui/download/DownloadViewModel.kt
```diff
@@ -170,9 +172,9 @@ class DownloadViewModel : ViewModel() {
 
         val values = currentDownloadsMutex.withLock {
             allValues.filter { card ->
-                val notImported = !card.isImported
+                val notImported = card.isImported != IMPORTED_PDF && card.isImported != IMPORT_SOURCE
                 val canDownload =
-                    card.downloadedTotal <= 0 || (card.downloadedCount * 100 / card.downloadedTotal) > 90
+                    card.downloadedTotal <= 0 || (card.downloadedCount * 100 / card.downloadedTotal) > 90// isn't better > 98? to be sure
```
LagradOst: No, 90 is good as it allows books to get many chapters at the same time. If you only have like 15 chapters and get 1 then 98% will not catch it.

app/src/main/java/com/lagradost/quicknovel/ui/download/AnyAdapter.kt
```diff
@@ -197,15 +199,19 @@ class AnyAdapter(
...
                             val diff = item.downloadedCount - epubSize
                             imageTextMore.text = "+$diff "
-                            imageTextMore.isVisible = diff > 0 && !showDownloadLoading && !item.isImported
+                            imageTextMore.isVisible = diff > 0 && !showDownloadLoading && item.isImported != IMPORT_SOURCE && item.isImported != IMPORTED_PDF
                             imageText.text = item.name
-                            imageView.setImage(item.image)
+
+                            imageView.alpha = if (isAPdfDownloading) 0.6f else 1.0f
+                            //avoid flickering
+                            if(!isAPdfDownloading || imageView.drawable == null) imageView.setImage(item.image)
```
LagradOst: 🤨 If the drawable is null then this makes the imageview use the img of something else. This should not be here impo.

app/src/main/java/com/lagradost/quicknovel/ReadActivity2.kt
```diff
@@ -1292,6 +1292,12 @@ class ReadActivity2 : AppCompatActivity(), ColorPickerDialogListener {
                 }
             }
 
+            //place it above binding.readApplyTranslation.setOnClickListener due to initialization reasons
```
LagradOst: ????

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1631,270 @@ object BookDownloader2 {
...
+        if (bitmap.isRecycled) return null
+
+        try {
+            val outStream = ByteArrayOutputStream()
+
+            //get image compressed as byteArray
+            bitmap.compress(Bitmap.CompressFormat.JPEG, 80, outStream)
+            val imgBytes = outStream.toByteArray()
+
+            // only over 90KB images
```
LagradOst: This comment is outdated.

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1631,270 @@ object BookDownloader2 {
...
+                        {
+                            changeDownload(id) {
+                                state = newState
+                            }?.let { progressState ->
+                                createNotification(id, load, progressState)
+                            }
+                            currentState = newState
+                        }
+                    }
+                    if(currentState == DownloadState.IsPaused) delay(1000)
```
LagradOst: This is a very long delay.

LagradOst: Use 200 like `downloadWorkThread`

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1631,270 @@ object BookDownloader2 {
...
+                                    this.progress = this.total
+                                    this.total = this.total
+                                }?.let { createNotification(id, load, it) }
+                            }
+                        }
+                        pageIdx++
+                    }
+
+                    //check notification options
+                    when (consumeAction(id)) {
```
LagradOst: This consumeAction is placed weird as it makes it possible to "pause" after you have finished, causing it to break entierly.

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1631,270 @@ object BookDownloader2 {
...
+    @Throws
+    suspend fun downloadPDFWorkThread(data: Uri, context: Context) {
+        //init pdf reader
+        if (!PDFBoxResourceLoader.isReady())
+            PDFBoxResourceLoader.init(context)
+        val filesDir = context.filesDir
+        val apiName = IMPORTED_PDF
+
+        // open file
+        context.contentResolver.openInputStream(data).use { inputStream ->
```
LagradOst: This creates a very nested structure, just use something like 
```kotlin
val document:PDDocument? = context.contentResolver.openInputStream(data).use { inputStream ->
    PDDocument.load(inputStream)
}
```

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1631,270 @@ object BookDownloader2 {
...
+                var currentState = DownloadState.IsDownloading
+                setPrefixData(load, apiName, totalPages.toLong(), 0L)
+                changeDownload(id) {
+                    state = currentState
+                    this.progress = 0L
+                    this.total = this.total
+                }?.let { createNotification(id, load, it) }
+
+                var pageIdx = 1
+                while (currentState != DownloadState.IsDone && currentState != DownloadState.IsStopped) {
```
LagradOst: This is very fragile, please use while true with breaks instead.

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1631,270 @@ object BookDownloader2 {
...
+        }
+        return resultResource
+    }
+
+
+
 
+    @WorkerThread
+    @Throws
+    suspend fun downloadPDFWorkThread(data: Uri, context: Context) {
```
LagradOst: No setSuffixData?

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1631,270 @@ object BookDownloader2 {
...
+    ): Resource? {
+        var resultResource: Resource? = null
+        val bitmap = imageXObject.image ?: return null
+        if (bitmap.isRecycled) return null
+
+        try {
+            val outStream = ByteArrayOutputStream()
+
+            //get image compressed as byteArray
+            bitmap.compress(Bitmap.CompressFormat.JPEG, 80, outStream)
```
LagradOst: Why JPEG and why not the best quality? Is PNG or WEBP_LOSSLESS not better?

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1631,270 @@ object BookDownloader2 {
...
+                        val bodyText = stripper.getTextForRegion("body")
+
+                        //if there's text, include
+                        if (bodyText.isNotBlank())
+                            currentChapterText.append(bodyText.textToHtmlChapter())
+
+                        //get images from page
+                        val resources = page.resources
+                        for (objName in resources.xObjectNames) {
+                            if (resources.isImageXObject(objName))
```
LagradOst: reverse the if to reduce nesting.

```kotlin
if (!resources.isImageXObject(objName)) continue
...
```

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1631,270 @@ object BookDownloader2 {
...
+                //notification
+                var currentState = DownloadState.IsDownloading
+                setPrefixData(load, apiName, totalPages.toLong(), 0L)
+                changeDownload(id) {
+                    state = currentState
+                    this.progress = 0L
+                    this.total = this.total
+                }?.let { createNotification(id, load, it) }
+
+                var pageIdx = 1
```
LagradOst: This is called Idx like it is an index, but is a number. It should start at 0 if you say it is an "Idx".

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1631,270 @@ object BookDownloader2 {
...
+                //to get text from pdf pages
+                val stripper = PDFTextStripperByArea()
+
+                //create sections of N pages
+                val pagesPerChapter = 10
+
+                //to create content of the chapter
+                val currentChapterText = StringBuilder()
+
+                var chapterCount = 1
```
LagradOst: Starts at 1?

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1631,270 @@ object BookDownloader2 {
...
+                var imageCount = 1
```
LagradOst: Starts at 1?

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -1431,6 +1449,16 @@ class ReadActivityViewModel : ViewModel() {
 
                     updateIndex(index)
 
+                    //preload next chapter
+                    viewModelScope.launch(Dispatchers.IO) {
+                        try {
+                            val exists = chapterMutex.withLock { chapterData[index + 1] is Resource.Success }
```
LagradOst: This exists check is useless as loadIndividualChapter already checks it.

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
TJMolina: It's because I don't like chapter 0; in a PDF, since there isn't an index that says what chapter it really is, it doesn't matter whether chapter 0 exists or not.

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
TJMolina: In this case it's because of an issue with EPUB indexes that I don't know how to solve; it absolutely has to start at 1.

TJMolina:
I also fixed a small bug that’s been in the app for a while. When you install the app for the first time, the settings default to grid mode, but the app actually shows everything in list mode by default. I just corrected the default in the settings and changed it to list mode so it makes sense

TJMolina:
These are imported PDFs, the progress is saved perfectly just as you had requested before:
image

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
TJMolina: Mm are you sure? TT it really hurts me to do that, I was forced to study structured programming... haha

TJMolina:
This is because I do NOT use the png or others

37MB for only a few images
image

this is with jpeg, 5MB
image

Moreover, an EPUB with heavy images takes a very long time to load with the current code for reading EPUBs

Moreover, JPEG compression does not affect it that much, and the loss of quality is hardly noticeable

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1647,328 @@ object BookDownloader2 {
...
+
+        //init pdf reader
+        if (!PDFBoxResourceLoader.isReady())
+            PDFBoxResourceLoader.init(context)
+
+        // open file
+        val document:PDDocument = context.contentResolver.openInputStream(data).use { inputStream ->
+            PDDocument.load(inputStream)
+        }?:run{
+            showToast(context.getString(R.string.failed), 1000)
```
LagradOst: I think you mean `showToast(R.string.failed, Toast.LENGTH_SHORT)` or `Toast.LENGTH_LONG`, as the duration is not in MS, nor do you have to get the string.

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1618,12 +1647,328 @@ object BookDownloader2 {
...
+                //check notification options
+                currentState = handleDownloadActions(id, load, currentState)
+                if(currentState == DownloadState.IsPaused)
+                {
+                    delay(200)
+                    continue
+                }
+                else if(currentState == DownloadState.IsStopped)
+                    break
+                else {
```
LagradOst: Remove the else here to reduce nesting.

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
TJMolina: "Haha, yeah yeah, I'll change it right away

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
TJMolina: I’ll make the small changes right away and I don’t have anything else to change. I have other things, but they’re for another pull request because they’re about adding Spanish. Or should I add everything in this pull request?

LagradOst: No, do not change the scope of this pull request. I merely mean that if you have any changes related to this pull request that you did felt was unfinished as normally the review-time may be long otherwise.

app/src/main/java/com/lagradost/quicknovel/ui/download/AnyAdapter.kt
```diff
@@ -190,7 +190,10 @@ class AnyAdapter(
                                     coverHeight
                                 )
                                 setOnClickListener {
-                                    downloadViewModel.readEpub(item)
+                                    if(item.downloadedCount != item.downloadedTotal)
+                                        downloadViewModel.refreshCard(item)
```
LagradOst: Wowowowow, what is going on here? You might still want to read an epub even if it is not full downloaded.

TJMolina: Yeah, I realized a second too late, I already fixed it in the commit minor details

TJMolina: The code for importing PDFs doesn’t generate the EPUB until it is fully scanned, so I added that if to resume the import when tapping the image if the process hasn’t finished yet

TJMolina: And I added the condition item.apiName == IMPORT_SOURCE_PDF

PR #403, opened 2026-01-25
Spanish + improvements

TJMolina:
I added Spanish language support to the app. I also increased the timeout for downloading the language model and for API requests, since they often fail on poor Wi-Fi connections. This doesn't affect normal usage at all. I also improved the regex for PDFs — it can now recognize letters from other languages and accented characters.


app/src/main/res/xml/settings.xml
```diff
@@ -117,7 +117,7 @@
 
     <PreferenceCategory
             android:key="extra"
-            android:title="Info"
+            android:title=""
```
LagradOst: ?

TJMolina: I don’t know what happened there, sorry, I’ll fix it.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -1104,7 +1099,7 @@ class ReadActivityViewModel : ViewModel() {
             mlTranslator?.closeQuietly()
             mlTranslator = null
 
-            if (settings.isInvalid() || settings.useOnlineTranslation) {
+            if (settings.isInvalid()) {
```
LagradOst: Why did you remove this???? You dont need to have the local model installed if you use online translation.

app/src/main/res/drawable/background_no_exist_novels.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
LagradOst: What is this file?

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: My mistake, I forgot that when I was experimenting.

app/src/main/res/drawable/background_no_exist_novels.xml
TJMolina: I was trying to add it but I changed my mind and forgot that resource 😅

LagradOst (approved):
Very good pr!

PR #406, opened 2026-02-06
reading progress v1

TJMolina:
I introduced a small indicator of your current reading progress, which also serves to identify if new chapters of the novel in question have been released.

progressReading

On the other hand, I unified the logic of the novelbin provider with that of allnovel because they were almost identical.

I also fixed small visual bugs regarding the update of the read/unread status of a chapter.

### DIFF
--- a/app/src/main/java/com/lagradost/quicknovel/providers/NovelBinProvider.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/providers/NovelBinProvider.kt
@@ -104,11 +104,12 @@ open class NovelBinProvider : AllNovelProvider() {
                 SearchResponse(
                     name = a.text(),
                     url = fixUrlNull(a.attr("href")) ?: return@mapNotNull null,
-                    fixUrlNull(element.selectFirst("div > div > img")?.attr("src")?.fullPosterFix()),
+                    fixUrlNull(element.selectFirst("div > div > img")?.attr("data-src")?.fullPosterFix()),
                     null,
                     null,
                     this.name
                 )
             })
     }
+
 }


app/src/main/java/com/lagradost/quicknovel/ui/download/DownloadViewModel.kt
```diff
@@ -618,4 +635,69 @@ class DownloadViewModel : ViewModel() {
...
         fetchAllData(true)
     }
+
+
+    private val loadingJobs = ConcurrentHashMap<Int, Job>()//15 jobs
+    private val downloadSemaphore = Semaphore(5)//from 15 jobs, only 5 at the same time
+    private val _chaptersUpdateSignal = MutableSharedFlow<Int>()
+    val chaptersUpdateSignal = _chaptersUpdateSignal.asSharedFlow()
+    val loadingStatus: MutableSet<Int> = Collections.newSetFromMap(ConcurrentHashMap<Int, Boolean>())
+    fun getReadingProgress(cached: ResultCached) {
```
LagradOst: This is a weird way to update reading progress, as it does it raw. Any apis used should use the APIRepository for cache and rate limiting. Also see downloadWorkThread as this does not handle key migration. Moreover, this sort of batched work should be done in the background and should be invoked from DownloadFileWorkManager.

LagradOst: Any network request done should be initiated by the user.

app/src/main/java/com/lagradost/quicknovel/ui/download/ViewpagerAdapter.kt
```diff
@@ -121,6 +127,15 @@ class ViewpagerAdapter(
                         footers = if(position == 0) 1 else 0
                         setHasStableIds(true)
                         submitList(item.items)
+                        fragment.viewLifecycleOwner.lifecycleScope.launch{
```
LagradOst: This should not be done on the view level, but in the viewmodel.

app/src/main/java/com/lagradost/quicknovel/ui/download/ViewpagerAdapter.kt
```diff
@@ -121,6 +127,15 @@ class ViewpagerAdapter(
...
+                            downloadViewModel.chaptersUpdateSignal.collectLatest { updateId->
+                                val currentItems = immutableCurrentList
+                                val index = currentItems.indexOfFirst { (it as? ResultCached)?.id == updateId }
+                                if (index != -1) {
+                                    notifyItemChanged(index)
```
LagradOst: Android uses the DiffCallback to update views, so this logic should not be done here.

app/src/main/java/com/lagradost/quicknovel/ui/download/ViewpagerAdapter.kt
```diff
@@ -130,22 +145,29 @@ class ViewpagerAdapter(
...
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
-                setOnScrollChangeListener { _, _, scrollY, _, oldScrollY ->
-                    val diff = scrollY - oldScrollY
-                    if (diff == 0) return@setOnScrollChangeListener
+                    val isCurrentlyScrolling = newState != androidx.recyclerview.widget.RecyclerView.SCROLL_STATE_IDLE
 
-                    scrollCallback.invoke(diff > 0)
+                    if (!isCurrentlyScrolling) {
+                        downloadViewModel.senDataToReadingProgress(adapter as? AnyAdapter, layoutManager as? LinearLayoutManager)
+                        adapter?.notifyDataSetChanged()
```
LagradOst: There should never be a call to notifyDataSetChanged.

app/src/main/java/com/lagradost/quicknovel/util/ResultCached.kt
LagradOst, line 9: This data should be immutable.

app/src/main/java/com/lagradost/quicknovel/ui/download/DownloadViewModel.kt
```diff
@@ -618,4 +635,69 @@ class DownloadViewModel : ViewModel() {
...
+                    loadingJobs.remove(oldestId)?.cancel()
+                    loadingStatus.remove(oldestId)
+                }
+            }
+        }
+
+        loadingJobs[id] = viewModelScope.launch(Dispatchers.IO) {
+            try {
+                downloadSemaphore.withPermit {
+                    val api = Apis.apis.find { it.name == cached.apiName }
```
LagradOst: getApiFromNameOrNull

LagradOst (changes requested):
Very cool feature, but it looks like you have a slight misunderstanding on how views get updated. The viewmodel holds all the data about all views, and uses postValue which is observed to update the recycleviews. The data should not be changed in-between observe invocations.

Furthermore it looks like you look for "new chapters" even if they are not downloaded. This makes many get requests without user intervention, slowing down the device and draining data. Please make it use the same "swipe" to refresh like downloads.

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1224,6 +1233,75 @@ object BookDownloader2 {
...
+    suspend fun getReadingProgress(cached: ResultCached)
+    {
+        downloadSemaphore.withPermit {
+            try
+            {
+                val api = getApiFromNameOrNull(cached.apiName) ?: return@withPermit
+                val response = api.load(cached.source, true)
+                if (response is com.lagradost.quicknovel.mvvm.Resource.Success)
+                {
+                    val loaded = response.value as StreamResponse
```
LagradOst: This will throw exceptions. Not every response is a StreamResponse.

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1224,6 +1233,75 @@ object BookDownloader2 {
...
+                                migrateKeys(oldId, newId, cached.name, loaded.name)
+                            }
+                        }
+                    }
+                }
+            } catch (e: Throwable) {
+                if (e !is CancellationException) logError(e)
+            }
+            finally {
+                readingProgressChanged.invoke(cached.id)
```
LagradOst: This should only be done in case of a successful refresh. Otherwise we waste time on updates.

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1224,6 +1233,75 @@ object BookDownloader2 {
...
+
+    private val downloadSemaphore = Semaphore(5)
+    suspend fun getReadingProgress(cached: ResultCached)
+    {
+        downloadSemaphore.withPermit {
+            try
+            {
+                val api = getApiFromNameOrNull(cached.apiName) ?: return@withPermit
+                val response = api.load(cached.source, true)
+                if (response is com.lagradost.quicknovel.mvvm.Resource.Success)
```
LagradOst: This can be inverted to ``` if (response !is Resource.Success) return``` to avoid unnecessary nesting.

app/src/main/java/com/lagradost/quicknovel/ui/result/ResultViewModel.kt
```diff
@@ -372,6 +376,16 @@ class ResultViewModel : ViewModel() {
         }
     }
 
+    fun getLastRead():String?{
+        val load = load as StreamResponse
```
LagradOst: This will throw an exception. Use `as? StreamResponse ?: return` instead.

app/src/main/java/com/lagradost/quicknovel/ui/result/ResultFragment.kt
```diff
@@ -112,6 +113,21 @@ class ResultFragment : Fragment() {
 
     override fun onResume() {
         super.onResume()
+        //only if really is onResume
+        if(viewModel.isResume){
```
LagradOst, line 119: Why only update the last chapters? Simply call refresh on all the views in the adapter with by the RecyclerView.ViewHolder.

app/src/main/java/com/lagradost/quicknovel/ui/download/DownloadViewModel.kt
```diff
@@ -513,7 +519,36 @@ class DownloadViewModel : ViewModel() {
...
+    fun debounceAction(actions: suspend()-> Unit){
+        debounceJob?.cancel()
+        debounceJob = viewModelScope.launch(Dispatchers.IO) {
+            delay(3000)
+            actions()
+        }
+    }
+    fun readingProgressChanged(tab: Int){
+        debounceAction {
+            loadAllData(false)
```
LagradOst: This is an expensive operation, for something that will be called a lot. I suggest making this into a surgical update where you use the information already within the cards to update it. Then you wont need a long 3s denounce too.

LagradOst (changes requested):
Very good improvements, however I am still a bit worried about performance and exceptions.

app/src/main/java/com/lagradost/quicknovel/ui/result/ResultFragment.kt
TJMolina: I didn't understand, do you mean I should use notifyDataSetChanged() in that part?

LagradOst: ... Sure, it is not the optimal solution, but given that it is only called once per resume it should be fine.

app/src/main/java/com/lagradost/quicknovel/ui/BaseAdapter.kt
```diff
@@ -235,7 +238,12 @@ abstract class BaseAdapter<
             CONTENT -> {
                 val realPosition = position - headers
                 val item = getItem(realPosition)
-                onUpdateContent(holder, item, realPosition)
+                if(payloads.contains(ONLY_PROGRESS_READING_TEXT)){
```
LagradOst: No, baseadapter should not contain any code related to quicknovel. Think of it as an independent lib.

app/src/main/java/com/lagradost/quicknovel/DataStore.kt
```diff
@@ -60,6 +60,8 @@ const val RESULT_BOOKMARK: String = "result_bookmarked"
 const val RESULT_BOOKMARK_STATE: String = "result_bookmarked_state"
 const val HISTORY_FOLDER: String = "result_history"
 const val CURRENT_TAB : String = "current_tab"
+const val EPUB_CURRENT_TOTAL_CHAPTERS = "epub_current_total_chapters"
```
LagradOst: Why did you add this? I do not see why this is needed.

TJMolina: CURRENT_TOTAL_CHAPTERS is really, really important for optimization. Since it’s not efficient to perform 500 submitList calls when using DiffCallback, I decided that every time the element goes through onBindContent, it should query the total number of chapters. That way, in BookDownload I only update that key, and the RecyclerView automatically checks it each time in onBindContent.

I also do it this way because each call to notifyItemChanged consumes a lot of resources, causing small stutters on the screen. For that reason, I only update the elements if they are visible, to reduce the number of calls to notifyItemChanged.

After that, totalChapters is updated exclusively through onBindContent when using get() for the key CURRENT_TOTAL_CHAPTERS.

TJMolina: notifyItemChanged doesn’t affect much if there are only a few elements, but if I update 500 elements at different times with short intervals, the stutters or slowdowns become really noticeable.

TJMolina: On the other hand, so far I haven’t noticed any slowdown or stutter when I use int get() = getKey().

app/src/main/java/com/lagradost/quicknovel/ui/BaseAdapter.kt
TJMolina: okk

app/src/main/java/com/lagradost/quicknovel/DataStore.kt
TJMolina: Or are setKey and getKey also expensive operations?

LagradOst: No, set and get is quite fast, the problem is that RESULT_BOOKMARK already contains this information, so this stores the same information twice.

TJMolina: So what do you think if I modify all the logic involving ResultCached so that it doesn’t use totalchapters, and instead stores that information in CURRENT_TOTAL_CHAPTERS?

I insist on using this new key because totalchapters, being part of the immutable element, the only way to tell the RecyclerView that it changed is to copy the entire list of pages, modify it, and re-upload it with submitList. But doing that 500 times randomly in short intervals is very inefficient. Likewise, calling notifyItemChanged 500 times in short bursts is also inefficient.

That’s why it’s better to use totalchapters: int get() = getKey. It’s easier because onBindContent automatically fetches the newest result every time the element is on screen.

LagradOst: I understand that android is slow af, but you can still use getKey<ResultCached>(RESULT_BOOKMARK, id).totalchapters instead of adding a new key.

TJMolina: You're a genius, man, I didn’t think of that haha, sometimes I don’t realize the simplest solutions TT

PR #412, opened 2026-02-19
Fix HTTP 403 error by adding User-Agent header

hamunoz:
Added User-Agent header to OkHttp requests to handle HTTP 403 errors.

coil3.network.HttpException: HTTP 403


app/src/main/java/com/lagradost/quicknovel/ui/download/AnyAdapter.kt
TJMolina: You noticed it before I did! haha, I don’t know how I missed it, my mistake.

PR #416, opened 2026-03-09
CloudflareKiller implemented in wtrlab

TJMolina:
In this PR I implemented the CloudflareKiller improvement (from a member of the Discord group) in wtrlab, creating an interceptor that can be used in the future by any provider that has Cloudflare.

### DIFF
--- a/app/src/main/java/com/lagradost/quicknovel/MainActivity.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/MainActivity.kt
@@ -51,6 +51,7 @@ import com.lagradost.quicknovel.mvvm.logError
 import com.lagradost.quicknovel.mvvm.observe
 import com.lagradost.quicknovel.mvvm.observeNullable
 import com.lagradost.quicknovel.mvvm.safe
+import com.lagradost.quicknovel.network.CloudflareKiller
 import com.lagradost.quicknovel.providers.RedditProvider
 import com.lagradost.quicknovel.ui.ReadType
 import com.lagradost.quicknovel.ui.download.DownloadFragment
@@ -111,7 +112,7 @@ class MainActivity : AppCompatActivity() {
             OkHttpClient()
                 .newBuilder()
                 .ignoreAllSSLErrors()
-                .readTimeout(50L, TimeUnit.SECONDS)//to online translations
+                .readTimeout(30L, TimeUnit.SECONDS)
                 .build(),
             responseParser = object : ResponseParser {
                 val mapper: ObjectMapper = jacksonObjectMapper().configure(
@@ -139,6 +140,16 @@ class MainActivity : AppCompatActivity() {
             defaultHeaders = mapOf("user-agent" to USER_AGENT)
         }
 
+        val appWithInterceptor by lazy {
+            Requests(
+                baseClient = app.baseClient.newBuilder()
+                    .addInterceptor(CloudflareKiller())
+                    .build(),
+                responseParser = app.responseParser
+            ).apply {
+                defaultHeaders = app.defaultHeaders
+            }
+        }
 
         // === API ===
         lateinit var navOptions: NavOptions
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/lagradost/quicknovel/MainActivity.kt
```diff
@@ -111,7 +112,7 @@ class MainActivity : AppCompatActivity() {
             OkHttpClient()
                 .newBuilder()
                 .ignoreAllSSLErrors()
-                .readTimeout(50L, TimeUnit.SECONDS)//to online translations
+                .readTimeout(30L, TimeUnit.SECONDS)
```
TJMolina, line 115: I forgot to change this, but it's better to keep it; in the next PR I greatly improved the online translator, increasing the speed by approximately 30%, so an excessive read timeout was unnecessary

app/src/main/java/com/lagradost/quicknovel/providers/ReadOnlineFreeBookProvider.kt
```diff
@@ -0,0 +1,72 @@
...
+        return "$title<br>$content"
+    }
+
+    override suspend fun search(query: String): List<SearchResponse> {
+        val document =
+                app.get("$mainUrl/index/search/q/${
+                    query.replace(" ", "%20")
+                }").document
+
+        return document.select("div.section-left div.capnhat div.section-bottom div").mapNotNull { parent ->
```
hamunoz, line 38: The search doubles the results 
change
```
document.select("div.section-left div.capnhat div.section-bottom div")
```
by
```
document.select("div.section-left div.capnhat div.section-bottom > div:nth-child(1)")
```

app/src/main/java/com/lagradost/quicknovel/providers/ReadOnlineFreeBookProvider.kt
```diff
@@ -0,0 +1,72 @@
...
+        val document = app.get(url).document
+        val intro = document.selectFirst("div.intro_novel")?:return null
+        val title = intro.selectFirst("div.title")?.html()?:""
+        val content = intro.selectFirst("div.content_novel")?.html()?:return null
+        return "$title<br>$content"
+    }
+
+    override suspend fun search(query: String): List<SearchResponse> {
+        val document =
+                app.get("$mainUrl/index/search/q/${
```
hamunoz, line 34:
```
app.get(
                "$mainUrl/index/search/q/${query.replace(" ", "%20")}",
                headers = baseHeaders  // ← Here we add the headers
            ).document
```

app/src/main/java/com/lagradost/quicknovel/providers/ReadOnlineFreeBookProvider.kt
hamunoz, line 1: has code error: 521 in search for certain novels
example: The Fellowship of the Ring
The error is also made directly from the website

app/src/main/java/com/lagradost/quicknovel/providers/ReadOnlineFreeBookProvider.kt
LagradOst: I fixed this

app/src/main/java/com/lagradost/quicknovel/providers/ReadOnlineFreeBookProvider.kt
LagradOst: Then the site is wrong, not sure what we can fix about that.

TJMolina: I'll check it

TJMolina: I already found out why it happens. When words other than the first start with a capital letter, it causes an internal error on the website. Try searching The fellowship of the ring instead of The Fellowship of the Ring. I'll fix it in the next PR, it's easy to solve

PR #418, opened 2026-03-14
Translations, epubs improvements and performance improved

TJMolina:
I grouped several changes into the same pull request to save time.

What did I do in this pull request?

I added ReadOnlineFreeBook to the manifest and the README.

When importing PDFs, it didn’t seem correct to call each part a "chapter" because they aren’t actually chapters, so I renamed them to "section", which seems more appropriate.

There was a bug where deleting a downloaded novel caused the app to crash if the deletion triggered a list reordering. I fixed it by changing the DiffCallback of the AnyAdapter to use id instead of source for downloaded novels.

I modified the logic of RegularBook so that images and sections are recognized correctly.

I also modified the logic of onScroll, updateIndex, and updateIndexAsync to avoid so many calls. These changes were made in ReadActivityViewModel and ReadActivity2.

I improved the online translation to send more text in different chunks. I also moved all the logic into a separate class.

I modified the code so that the text-to-speech reader does not read image alt text, since it is usually not very descriptive and was annoying.

I also fixed an issue where images didn’t have scroll support. Because of that, if you finished reading while viewing an image, when you returned the scroll position would be somewhere else.

### DIFF
--- a/app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
@@ -364,7 +364,7 @@ class RegularBook(val data: EpubBook) : AbstractBook() {
             try {
                 val ref = spine.spineReferences[i]
 
-                // Si el recurso no es lineal y no es el inicio explícito de este capítulo, saltar
+                // I have no idea, but nonlinear = stop?
                 if (!ref.isLinear && i != startIdx) continue
 
                 val html = ref.resource.reader.readText()
@@ -965,12 +965,17 @@ class ReadActivityViewModel : ViewModel() {
             val textHash = hashString(
                 text.trim().toString().toByteArray()
             )
-            val filePrefix = "ml_${textHash}.${currentSettings.from}_to_${currentSettings.to}.${
+
+            //the file
+            val filePrefix =
+                "ml_${textHash}.${currentSettings.from}_to_${currentSettings.to}.${
                 if (currentSettings.useOnlineTranslation) "online" 
                 else "offline"
             }"
 
-            //Try to use cache
+            // read from cache if it exists
+            // we assume that parseTextToSpans is equivalent from restoring from the builder
+            // aka out == parseTextToSpans(builder)
             val cachedData = safe {
                 context?.cacheDir?.let { dir ->
                     val cache = File(dir, "$filePrefix.txt")
--- a/app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
@@ -293,8 +293,8 @@ class QuickBook(val data: QuickStreamData) : AbstractBook() {
     }
 }
 
-class RegularBook(val data: EpubBook) : AbstractBook() {private val allTocReferences: List<TOCReference>
-
+class RegularBook(val data: EpubBook) : AbstractBook() {
+    private val allTocReferences: List<TOCReference>
     init {
         val flatTOC = mutableListOf<TOCReference>()
         fun flatten(refs: List<TOCReference>) {
@@ -609,24 +609,22 @@ class ReadActivityViewModel : ViewModel() {
         val range =
             if(!notify) (index - initPaddingBottom..index + initPaddingTop)
             else (index - chapterPaddingBottom..index + chapterPaddingTop)
+
         for (idx in range) {
-            requested += index
+            if(requested.contains(idx)) continue
+            requested += idx
             loadIndividualChapter(idx, notify = notify, postLoading = postLoading)
         }
 
     }
 
     private fun updateIndex(index: Int) {
-        var alreadyRequested = false
-        if (!requested.contains(index)) {
-            alreadyRequested = true
-        }
-        requested += index
-        if (alreadyRequested) return
-
-        ioSafe {
-            updateIndexAsync(index)
-        }
+        val range = (index - chapterPaddingBottom .. index + chapterPaddingTop)
+        val needToLoad = range.any { !requested.contains(it) }
+        if (needToLoad)
+            ioSafe {
+                updateIndexAsync(index)
+            }
     }
 
     fun onScroll(visibility: ScrollVisibilityIndex?) {
@@ -935,6 +933,26 @@ class ReadActivityViewModel : ViewModel() {
         return sb.toString()
     }
 
+    private fun getFinalTranslatedText(spans: ArrayList<TextSpan>, translatedLines: List<String>):Pair<SpannableStringBuilder, ArrayList<TextSpan>>{
+        val builder = SpannableStringBuilder()
+        val out = ArrayList<TextSpan>()
+        spans.forEachIndexed { i, originalSpan ->
+            val hasImage = originalSpan.text.getSpans<AsyncDrawableSpan>().isNotEmpty()
+            val finalText =
+                if (hasImage)
+                    originalSpan.text
+                else
+                    (translatedLines.getOrNull(i)?: return@forEachIndexed).toSpanned()
+            val start = builder.length
+            builder.append(finalText)
+            val end = builder.length
+            builder.append('\n')
+            out.add(TextSpan(finalText, start, end, originalSpan.index, originalSpan.innerIndex))
+        }
+        return builder to out
+    }
+
+
     @Throws(MLException::class)
     private suspend fun translate(
         text: Spanned,
@@ -944,8 +962,6 @@ class ReadActivityViewModel : ViewModel() {
         try {
             val currentSettings = mlSettings
             if (spans.isEmpty() || currentSettings.isInvalid()) return text to spans
-            val builder = SpannableStringBuilder()
-            val out = ArrayList<TextSpan>()
             val textHash = hashString(
                 text.trim().toString().toByteArray()
             )
@@ -961,19 +977,7 @@ class ReadActivityViewModel : ViewModel() {
                     if (cache.exists()) {
                         Log.i(TAG, "Cache exists for $filePrefix")
                         val lines = cache.readLines()
-                        spans.forEachIndexed { i, originalSpan ->
-                            val hasImage = originalSpan.text.getSpans<AsyncDrawableSpan>().isNotEmpty()
-                            val finalText =
-                                if (hasImage)
-                                    originalSpan.text
-                                else
-                                    lines[i].toSpanned()
-                            val start = builder.length
-                            builder.append(finalText)
-                            val end = builder.length
-                            builder.append('\n')
-                            out.add(TextSpan(finalText, start, end, originalSpan.index, originalSpan.innerIndex))
-                        }
+                        val (builder, out) = getFinalTranslatedText(spans, lines)
                         return@safe builder to out
                     }
                 }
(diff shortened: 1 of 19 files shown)


app/src/main/java/com/lagradost/quicknovel/ui/download/AnyAdapter.kt
```diff
@@ -39,7 +39,7 @@ class AnyAdapter(
             if (a is ResultCached && b is ResultCached) {
                 a.source == b.source
             } else if (a is DownloadFragment.DownloadDataLoaded && b is DownloadFragment.DownloadDataLoaded) {
-                a.source == b.source
+                a.id == b.id//if you use .source, when you delete a novel, this crashes
```
LagradOst, line 42: Ye, this should be related to the import EPUB where it has no source triggering some IllegalStateException or similar.

app/src/main/java/com/lagradost/quicknovel/ui/TextAdapter.kt
```diff
@@ -414,35 +414,55 @@ class TextAdapter(
...
-            if (span !is TextSpan) return emptyList()
-
-            val outLocation = IntArray(2)
-            binding.root.getLocationInWindow(outLocation)
-            val y = outLocation[1] + binding.root.paddingTop
-
-            //val paddingTop =
-            //val paddingBottom = binding.root.paddingBottom
 
+            val binding = viewHolder.view
```
LagradOst, line 418: Nice!

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -593,28 +606,51 @@ class ReadActivityViewModel : ViewModel() {
...
     }
 
     private fun updateIndex(index: Int) {
         var alreadyRequested = false
-        for (idx in index - chapterPaddingBottom..index + chapterPaddingTop) {
-            if (!requested.contains(index)) {
-                alreadyRequested = true
-            }
-            requested += index
+        if (!requested.contains(index)) {
```
LagradOst: This is slightly wrong. Well the old code was wrong too, but you should fix it by checking if all are requested, and only return then.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -907,111 +943,106 @@ class ReadActivityViewModel : ViewModel() {
...
+            //Try to use cache
+            val cachedData = safe {
+                context?.cacheDir?.let { dir ->
+                    val cache = File(dir, "$filePrefix.txt")
                     if (cache.exists()) {
                         Log.i(TAG, "Cache exists for $filePrefix")
-                        val mlText = cache.readText().toSpanned()
-                        return@safe mlText to parseTextToSpans(mlText, spans[0].index)
+                        val lines = cache.readLines()
+                        spans.forEachIndexed { i, originalSpan ->
```
LagradOst: It looks like you are doing the same thing thrice here, try to make it a function.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -907,111 +943,106 @@ class ReadActivityViewModel : ViewModel() {
...
-                            spans[i].innerIndex
-                        )
-                    )
+                    out.add(TextSpan(finalText, start, end, span.index, span.innerIndex))
                 }
             }
-            val mlRawText = builder.toString()
 
-            // atomically write the file by rename
+            //Save in cache
```
LagradOst: Do not change comments.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -907,111 +943,106 @@ class ReadActivityViewModel : ViewModel() {
...
             safe {
-                context?.cacheDir?.let {
-                    val cache = File(it, "$filePrefix.tmp")
-                    cache.writeText(mlRawText)
-                    safe { File(it, "$filePrefix.txt").delete() } // just in case
-                    cache.renameTo(File(it, "$filePrefix.txt"))
+                context?.cacheDir?.let { dir ->
+                    val cache = File(dir, "$filePrefix.tmp")
+                    cache.writeText(builder.toString())
+                    val finalFile = File(dir, "$filePrefix.txt")
```
LagradOst: Why did you remove 1. comments 2. the inner safe?

LagradOst (changes requested):
Very good changes. Just please do not remove/update/change comments.

app/src/main/res/values-es/strings.xml
hamunoz: search
```    
<string name="conection_error">Error de conección.</string> 
```
change
``` 
    <string name="conection_error">Error de conexión.</string>
```

TJMolina: Haha, I'm so silly.

app/src/main/java/com/lagradost/quicknovel/util/GoogleTranslateOnline.kt
```diff
@@ -0,0 +1,137 @@
...
+import com.fasterxml.jackson.annotation.JsonFormat
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.lagradost.quicknovel.MainActivity
+import kotlin.math.pow
+import android.net.Uri
+import com.lagradost.quicknovel.mvvm.logError
+import kotlinx.coroutines.delay
+
+
+class GoogleTranslateOnline(val loading: suspend (Int, Int) -> Unit = {_,_->})
```
LagradOst: There is no reason for this to be a class when it does not have state. 

I recommend refactoring it into
 `object GoogleTranslateOnline {` 
and 

```kotlin
 suspend fun onlineTranslate(
        paragraphs: List<String>,
        from: String,
        to: String,
        loading: suspend (Int, Int) -> Unit = { _, _ -> }
    ): List<String> {
```

app/src/main/java/com/lagradost/quicknovel/util/GoogleTranslateOnline.kt
```diff
@@ -0,0 +1,137 @@
...
+    @JsonFormat(shape = JsonFormat.Shape.ARRAY)
+    data class GoogleSentence(
+        val trans: String,
+        val orig: String,
+        val translit: String? = null,
+        val srcTranslit: String? = null
+    )
+
+    private val baseUrl = "https://translate.googleapis.com/translate_a/single?client=gtx&sl="
+    val paragraphsSeparator ="\nXQZX\n"
```
LagradOst: const?

app/src/main/java/com/lagradost/quicknovel/util/GoogleTranslateOnline.kt
```diff
@@ -0,0 +1,137 @@
...
+    val paragraphsSeparatorRegex = Regex("\\n?XQZX\\n?")
+    val charsLimit = 2000
```
LagradOst: const?

app/src/main/java/com/lagradost/quicknovel/util/GoogleTranslateOnline.kt
```diff
@@ -0,0 +1,137 @@
...
+        val translatedChunks = chunks.mapIndexed { i, chunk ->
+            loading.invoke(i, chunks.size)
+            translateChunk(chunk, from, to)
+        }
+
+        return translatedChunks.flatMap { it.split(paragraphsSeparatorRegex) }
+    }
+
+
+    private suspend fun translateChunk(text: String, from: String, to: String): String {
```
LagradOst: I do not really understand why you have two functions that do the same.

translateChunk and retrySpecificSentence should be a single function.

app/src/main/java/com/lagradost/quicknovel/util/GoogleTranslateOnline.kt
```diff
@@ -0,0 +1,137 @@
...
+                if (retryNumber >= maxRetry){
+                    return text
+                }
+                delay(500L * (2.0.pow(retryNumber).toLong()))
+            }
+        }
+        return text
+    }
+
+    fun List<String>.chunkByLimit(): List<String> {
```
LagradOst: This does not preserve the paragraphs correctly. Do not cut a single paragraph into several parts as that introduces paragraphs when doing translatedChunks.flatMap.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -297,108 +294,119 @@ class QuickBook(val data: QuickStreamData) : AbstractBook() {
...
-        if (endIdx == -1) {
-            endIdx = data.spine.size()
-        }
         val builder = StringBuilder()
-        for (i in startIdx until endIdx) {
-            try//this is for corrupted epubs like from annasarchive
-            {
-                val ref = data.spine.spineReferences[i]
+
+        val endIdx = if (nextChapterStartIdx < startIdx) startIdx + 1 else nextChapterStartIdx
```
LagradOst: What is wrong with the current code? I do not understand this change.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -297,108 +294,119 @@ class QuickBook(val data: QuickStreamData) : AbstractBook() {
...
+        val actualEndIdx = maxOf(startIdx + 1, endIdx)
+
+        for (i in startIdx until actualEndIdx) {
+            if (i >= spine.spineReferences.size) break
```
LagradOst: This should be moved into the variable, eg `minOf(actualEndIdx, spine.spineReferences.size)`

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -297,108 +294,119 @@ class QuickBook(val data: QuickStreamData) : AbstractBook() {
...
-        return false
-    }
+    private fun resolveRelativePath(basePath: String, relativePath: String): String {
+        val cleanRelative = relativePath.substringBefore("?").substringBefore("#")
+        val fullPath = if (basePath.isEmpty()) cleanRelative else "$basePath/$cleanRelative"
 
-    override suspend fun posterBytes(): ByteArray? {
-        return data.coverImage?.data
+        val parts = fullPath.split("/")
+        val resolvedParts = mutableListOf<String>()
```
LagradOst: ArrayDeque for addLast + removeLast

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -593,26 +606,47 @@ class ReadActivityViewModel : ViewModel() {
...
     ) {
-        for (idx in index - chapterPaddingBottom..index + chapterPaddingTop) {
-            requested += index
-            loadIndividualChapter(idx, reload = false, notify = notify, postLoading = postLoading)
+        val range =
+            if(!notify) (index - initPaddingBottom..index + initPaddingTop)
+            else (index - chapterPaddingBottom..index + chapterPaddingTop)
+
+        for (idx in range) {
+            if(requested.contains(idx)) continue
```
LagradOst: No, you cant do a continue here as requested is only used for updateIndex.

app/src/main/java/com/lagradost/quicknovel/util/GoogleTranslateOnline.kt
```diff
@@ -0,0 +1,137 @@
...
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    @JsonFormat(shape = JsonFormat.Shape.ARRAY)
+    data class GoogleSentence(
+        val trans: String,
+        val orig: String,
+        val translit: String? = null,
+        val srcTranslit: String? = null
+    )
+
+    private val baseUrl = "https://translate.googleapis.com/translate_a/single?client=gtx&sl="
```
LagradOst: const?

LagradOst (changes requested):
I have been very busy so this is a bit of a late review, but I have done a much more thorough review analyzing the behavior of the code. The code is good, but I have some comments about the GoogleTranslateOnline implementation.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
TJMolina: That part does exactly the same as the original; I just made it “cleaner” (or nicer) in my opinion. Do you want me to put it back the way it was?

app/src/main/java/com/lagradost/quicknovel/util/GoogleTranslateOnline.kt
TJMolina: If you’re referring to the part where I use the String.chunkByLimit() function, you’re right. But if you mean the entire List<String>.chunkByLimit() function, that doesn’t happen. I added a separator that I use to correctly identify each original paragraph. After running several tests, I found a separator that Google Translate doesn’t remove or ignore when merging paragraphs. So it would be very, very rare for there to be more paragraphs than the original, and also very, very rare—almost impossible—for paragraphs to be merged.

I also added a safeguard in ReadActivityViewModel so that all paragraphs are still displayed in the rare case that there are fewer, so it doesn’t really affect anything.

LagradOst: Ye, I was referring to the extra paragraphs caused by `String.chunkByLimit()`. Here is the example, that looks to be fixed by your latest commit. 

```
const val charsLimit = 3
print(listOf("1.1 1.2 1.3 1.4","2","3").chunkByLimit())
```
```
Incorrect output=
[1.1, 1.2, 1.3, 1.4
XQZX
, 2
XQZX
, 3
XQZX
]

Correct output=[1.1 1.2 1.3 1.4, 2, 3]
```

Maybe add a comment about this, as we never want to introduce or remove paragraphs.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
LagradOst: Please put it back, as the new code is more complex, at least impo.

app/src/main/java/com/lagradost/quicknovel/util/GoogleTranslateOnline.kt
TJMolina: Yes, I’ve already fixed it in the new commits.

LagradOst (approved):

PR #423, opened 2026-03-16
Llm assistant

Kvyii:
# LLM Base Language Assistant 

Adds an OpenRouter powered LLM translator for the purposes of language education.

cassie


TJMolina:
Personally, I liked your idea for implementing API keys. It could be used for the API that returns images, making it possible to give books custom covers, or to use the Gemini API key to translate books with AI like WTRLab does. It would be great if you could implement it.

Kvyii:
@TJMolina 
> Personally, I liked your idea for implementing API keys. It could be used for the API that returns images, making it possible to give books custom covers, or to use the Gemini API key to translate books with AI like WTRLab does. It would be great if you could implement it.

My apologies. I was unaware that forks default to the main branch of the original repo. The original purpose of this was to add an OpenRouter powered translator for my english learnings.

I don't believe it fits in with the spirit of the original QuickNovel which is by the original Author's definition "Adfree FOSS Android app for downloading novels". I suspect the majority of the user base has none or limited need for english <> english translation. Also the LLM functionality requires paid OpenRouter credits, an understanding of model selection / cost per token / output tokens pers second / combined with freely editable user prompt which can easily be broken through user error.

I've updated the PR body with some images of what the fork does. If the original author of this app likes, this I am happy to implement this as part of the main app instead of continuing down my fork :-)

PR #429, opened 2026-03-27
Add GitHub Actions workflow for building app

hamunoz:
This workflow builds the application and validates the Gradle wrapper on every push.


LagradOst:
Why was this closed?

hamunoz:
The action it creates doesn't work

PR #445, opened 2026-05-01
Some more providers

TJMolina:
I added some new providers that I didn’t get to include in the previous PR. I also noticed that people often don’t understand how the translation option works, so I added a label that might help alleviate that issue.


TJMolina:
Based on AbdullahM0hamed’s PR, I’ve completed the Wattpad provider.

app/src/main/java/com/lagradost/quicknovel/providers/WattpadProvider.kt
```diff
@@ -1,241 +1,228 @@
...
+                )
+            }
+    }
+
+    override suspend fun load(url: String): LoadResponse {
+        val response = app.get(url)
+        val script_regex = Regex("<script>.*?window._+remix.*?<\\/script>")
+        val script_text = script_regex.find(response.text)!!.value
+        val json_text = Regex("\\{.*\\}").find(script_text)!!.value
+        val wattpad_data = JSONObject(json_text)
```
LagradOst: Why are you using json objects intead of parsing the json with `parseJson`.

app/src/main/java/com/lagradost/quicknovel/providers/WattpadProvider.kt
```diff
@@ -1,241 +1,228 @@
...
 
-    )
+        val fullHtml = StringBuilder()
+
+        for (page in 1..totalPages) {
+            val pageUrl = "$textUrlBase$page"
+            val pageResp = app.get(pageUrl).text
+            fullHtml.append(pageResp)
+        }
+        return Jsoup.parse(fullHtml.toString()).html()
```
LagradOst: Why are you parsing then taking the html again? This does nothing?

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -568,7 +568,7 @@ object BookDownloader2Helper {
 
         val externalReader = settingsManager.getBoolean(
```
LagradOst, line 569: Wow, why are you changing the default? This should be a separate pull request with real motivation behind this change.

app/src/main/res/layout/read_bottom_settings.xml
```diff
@@ -502,6 +501,13 @@
                     app:icon="@drawable/ic_baseline_menu_book_24"
                     tools:text="English" />
 
+                <TextView
```
LagradOst, line 491: This might be better to do in https://github.com/LagradOst/QuickNovel/pull/454

app/src/main/java/com/lagradost/quicknovel/providers/LnMTLProvider.kt
```diff
@@ -0,0 +1,223 @@
...
+class LnMTLProvider : MainAPI() {
+
+    override val name = "LnMTL"
+    override val mainUrl = "https://lnmtl.com"
+    override val hasMainPage = true
+    override val iconId = R.drawable.icon_lnmtl
+    override val iconBackgroundId = R.color.white
+    private var allNovels: Array<NovelInfo> = emptyArray()
+
+    private suspend fun prefetchAllNovels(){
```
LagradOst: I think this should have a mutex lock or similar to prevent `prefetchAllNovels` to be called twice when one is loading.

LagradOst (changes requested):
Good changes! Only some minor code quality changes and stuff to be moved into another pull request, otherwise it looks great! ❤️ 

I will test this pull request more in depth after these small changes.

app/src/main/java/com/lagradost/quicknovel/providers/WattpadProvider.kt
TJMolina: Umm, it’s from the original code. I forgot to change it because it was working haha.

TJMolina:
I think I've finished all the modifications, but just today novlove changed its domain and entire structure, so I'll have to redo it from scratch 😭

TJMolina:
done

LagradOst (approved):
Great work!

PR #447, opened 2026-05-07
Russian localization

Xapitonov:
_**Added Russian localization to the project.**_

### DIFF
--- a/app/src/main/res/values-ru-rRU/strings.xml
+++ b/app/src/main/res/values-ru-rRU/strings.xml
@@ -1,0 +1,298 @@
+<resources>
+    <string name="app_name">QuickNovel</string>
+    <string name="title_search">Поиск</string>
+    <string name="title_history">История</string>
+    <string name="title_settings">Сервис</string>
+    <string name="title_home">Обзор</string>
+    <string name="title_download">Книги</string>
+
+    <string name="locale_key" translatable="false">locale_key</string>
+    <string name="provider_lang_key" translatable="false">provider_lang_key</string>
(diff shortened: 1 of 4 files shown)


LagradOst (changes requested):
Very cool change, but do not forget to also add it here https://github.com/LagradOst/QuickNovel/blob/master/app/src/main/java/com/lagradost/quicknovel/ui/settings/SettingsFragment.kt#L85C13-L85C26

Otherwise you will not be able to change the language from the app!

app/src/main/res/values-ru-rRU/strings.xml
```diff
@@ -0,0 +1,298 @@
+<resources>
```
hamunoz, line 1: You need to add in  
app/src/main/java/com/lagradost/quicknovel/ui/settings/SettingsFragment.kt
Specifically in: 
```kotlin
   val appLanguages = arrayListOf(
            /* begin language list */
            Triple("", "English", "en"),
            Triple("", "Türkçe", "tr"),
            Triple("", "Español", "es"),
            /* end language list */
        ).sortedBy { it.second.lowercase() } //ye, we go alphabetical, so ppl don't put their lang on top

```
Furthermore, to continue with the QuickNovel format, it should be the folder  values-ru

Xapitonov:
**Thank you so much for making these fixes yourself!**

_I wasn't able to do it myself because my smartphone broke and was in repair, but they never fixed it. 
Also, I had some family circumstances.
I really appreciate you taking the time to fix this._

**Thank you!**

PR #453, opened 2026-05-12
More tags/librarys

TJMolina:
This PR adds the functionality to create new sections/libraries/bookmarks, or whatever they’re called. They can be created, edited, and deleted. Their display order can also be changed—all for the convenience of even the most demanding users.


app/src/main/java/com/lagradost/quicknovel/DataStore.kt
```diff
@@ -220,4 +224,123 @@ object DataStore {
...
+const val LIBRARIES_KEY: String = "default_libraries"
+data class DefaultLibrary(
+    val id: Int,
+    val key: String,
+    val title: String,
+    val editable: Boolean = true,
+    val position: Int = 0
+)
+
+val DEFAULT_LIBRARIES: List<DefaultLibrary> = listOf(
```
LagradOst: Stringres like type_plan_to_read?

app/src/main/java/com/lagradost/quicknovel/ui/library/LibrarySectionAdapter.kt
```diff
@@ -0,0 +1,115 @@
...
+    override fun onBindContent(holder: ViewHolderState<Any>, item: DefaultLibrary, position: Int) {
+        val binding = holder.view as? ItemLibrarySectionBinding ?: return
+        val isBuiltIn = item.key in builtInKeys
+
+        binding.sectionName.text = item.title
+        binding.novelCount.text = "${counts[item.id] ?: 0}"
+
+        val bgColor =MaterialColors.getColor(binding.root, R.attr.colorPrimary, 0)
+        val fgColor = MaterialColors.getColor(binding.root, MatR.attr.colorOnSecondaryContainer, 0)
+        ViewCompat.setBackgroundTintList(binding.novelCount, ColorStateList.valueOf(bgColor))
```
LagradOst: What? Why are you doing this in code and not xml?

app/src/main/java/com/lagradost/quicknovel/ui/library/LibrarySectionFragment.kt
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+
+    override fun onDestroyView() {
+        _binding = null
+        super.onDestroyView()
+    }
+
+    private fun refresh() {
+        val ctx = requireContext()
```
LagradOst: Never use requireContext, it will crash the app.

app/src/main/java/com/lagradost/quicknovel/ui/library/LibrarySectionFragment.kt
```diff
@@ -0,0 +1,198 @@
...
+import com.lagradost.quicknovel.databinding.FragmentLibrarySectionBinding
+import com.lagradost.quicknovel.deleteLibrary
+import com.lagradost.quicknovel.getLibraries
+import com.lagradost.quicknovel.getLibraryBookmarkCount
+import com.lagradost.quicknovel.mergeLibraries
+import com.lagradost.quicknovel.mvvm.logError
+import com.lagradost.quicknovel.saveLibraries
+import com.lagradost.quicknovel.updateLibrary
+
+class LibrarySectionFragment : Fragment() {
```
LagradOst: BaseFragment

app/src/main/java/com/lagradost/quicknovel/ui/library/LibrarySectionAdapter.kt
```diff
@@ -0,0 +1,115 @@
...
+import com.lagradost.quicknovel.ui.ViewHolderState
+import java.util.Collections
+import com.google.android.material.R as MatR
+
+class LibrarySectionAdapter(
+    private val onRenameClick: (DefaultLibrary) -> Unit,
+    private val onMergeClick: (DefaultLibrary) -> Unit,
+    private val onDeleteClick: (DefaultLibrary) -> Unit,
+    private val onDragFinished: () -> Unit,
+) : BaseAdapter<DefaultLibrary, Any>(
```
LagradOst: NoStateAdapter

LagradOst (changes requested):
While working, the current design does not adhere to the general design of the app. The colors are slightly off, missing some animations, and the icons chosen is wrong. E.g. the merge icon is of refresh, not merge/union. You show a toast with "Done"? The back button on the top bar is not working. 

I am also unsure how robust this system is to backups, given that a backup overrides all keys, including the LIBRARIES_KEY, meaning that a backup may delete categories instead of adding to them.

TJMolina:
You’re absolutely right about the design haha, I’m terrible at it. Good thing I don’t work as a designer. I’ll fix the issues, but I’d need some help knowing how you’d like me to style it.

TJMolina:
About the merge/combine icon, I’m having a really hard time choosing one. Temporarily, I picked `ic_baseline_collections_bookmark_24`, but my other options are:

* `ic_baseline_code`
* `ic_baseline_open_in_new_24`
* `ic_baseline_tune_24`

TJMolina:
Im still fixing some errors

TJMolina:
Done! (I think haha). The only thing left is the styling part. I don’t have the same artistic vision as you haha, so if you guide me, I can make it however you think is best because I honestly have no idea 😅

TJMolina:
What do you think about this interface? It would replace the entire previous fragment. Or do you think it doesn't fit either?

image
image

TJMolina:
By the way, I have other pull requests currently in development, and they also involve UI changes. Can I send the screenshots on Discord to ask for feedback there instead of creating a pull request? That way I can avoid the problem of fully developing a UI that I might end up having to throw away anyway... haha. Or would you prefer that I only use GitHub?

LagradOst:
> By the way, I have other pull requests currently in development, and they also involve UI changes. Can I send the screenshots on Discord to ask for feedback there instead of creating a pull request? That way I can avoid the problem of fully developing a UI that I might end up having to throw away anyway... haha. Or would you prefer that I only use GitHub?

Yes that is a good point, use discord for fast answers

LagradOst:
> What do you think about this interface? It would replace the entire previous fragment. Or do you think it doesn't fit either?
> image image

That UI is fitting, but try to avoid situations where it is easy to misclick, or lessen it with popups. For example it might be better to add the "delete" button when editing the category instead of adding it as a button that is always shown. I have not tested the UI, but I would keep the current look we have right now and only show the "new" ui you have shown if the user press some edit button, as a user should rarely have to see all this information like reorder, rename ect.

TJMolina:
I'm going to close/delete this PR and open a new one. I'm so bad with Git that I just made a mess! But I'll leave a quick note explaining how this PR works:

Library base refactoring: I modified the library bases to make future changes easier, such as novel migration.

User customization: Added the option for users to customize libraries to their liking—add, delete, reorder, and rename.

In essence, it's just that; not much was actually changed. It might look like a lot of changes due to the Git mess.

This PR uses XML for the library dialog because MainActivity and ResultFragment (where the dialog is displayed) are still using XML.

I highly recommend merging this PR because it lays the groundwork for two major upcoming updates: Novels Update (notifying users when a new chapter is released) and Migrate Novels (moving a book to a different provider), plus whatever else comes next. The Novels Update feature is almost finished and depends on parts of this PR, so I can't submit it until this one is merged or reviewed.

PR #454, opened 2026-05-12
Modification of the toolbar in ReadingActivity2.kt

TJMolina:
In this PR, the options menu of ReadingActivity2 was modified because the original one was becoming larger and larger, making it very uncomfortable and inefficient to scroll so much just to find what you want. Because of this, the options were divided into 4 sections: Screen, Translation, Voice, and Text. A TabLayout was used for this.


LagradOst (approved):
Code looks good, will merge if no issues are found when testing!

LagradOst (changes requested):
Please use the correct theme for the tablayout. Check out how they look in the results page and the chapter dialog.

TJMolina:
image
like this?

LagradOst:
> image like this?

Yes exactly!

TJMolina:
I forgot to add the 'to' label, I'll do it tomorrow

TJMolina:
You might ask yourself, why use exactly the same arrow as the one above? It’s very important! A lot of people are lazy, distracted, or a bit silly (no offense haha), so it’s better to put it directly in their face.

LagradOst (approved):
Absolutely fantastic work!

PR #459, opened 2026-05-27
Minor fixes and Auto detect Languague

TJMolina:
Minor changes: fixed the WTRLab selector, added a rate limit to NovelFire, and added automatic language detection.


app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -2023,13 +2025,12 @@ class ReadActivityViewModel : ViewModel() {
...
             val all = TranslateLanguage.getAllLanguages()
 
-            if (!all.contains(to)) {
+            if (!all.contains(to) || to == AUTO_LANG) {
                 // no translation
                 return false
             }
 
-            if (!all.contains(from)) {
-                // no support for auto yet, see https://developers.google.com/ml-kit/language/identification/android
```
LagradOst, line 2032: Please do not remove this comment, as it links to the docs.

app/src/main/java/com/lagradost/quicknovel/ReadActivity2.kt
```diff
@@ -1333,11 +1334,23 @@ class ReadActivity2 : AppCompatActivity(), ColorPickerDialogListener {
...
+            binding.readOnlineTranslationSwitch.isChecked = viewModel.mlUseOnlineTransaltion
+            binding.readOnlineTranslationSwitch.setOnCheckedChangeListener { _, isChecked ->
+                viewModel.mlUseOnlineTransaltion = isChecked
+                if (!isChecked && viewModel.mlFromLanguage == ReadActivityViewModel.MLSettings.AUTO_LANG) {
+                    viewModel.mlFromLanguage = "en"
+                    binding.readMlFrom.text = ReadActivityViewModel.MLSettings.fromShortToDisplay("en")
+                }
+            }
+
             binding.readMlFrom.setOnClickListener { view ->
```
LagradOst, line 1348: Is this correct, or am I misreading this? readMlFrom can be auto sure, but readMlTo should never be auto correct?

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -2023,13 +2025,12 @@ class ReadActivityViewModel : ViewModel() {
...
+            if (!all.contains(from)  && !useOnlineTranslation) {
```
LagradOst: Unsure about this logic. Please use == AUTO_LANG

LagradOst (changes requested):
Code looks good, but I have not tested it yet.

app/src/main/java/com/lagradost/quicknovel/ReadActivity2.kt
TJMolina: yep, should never be auto_lang

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -2023,13 +2025,17 @@ class ReadActivityViewModel : ViewModel() {
...
                 // no translation
                 return false
             }
 
-            if (!all.contains(from)) {
-                // no support for auto yet, see https://developers.google.com/ml-kit/language/identification/android
+            // no support for auto yet (for offlineTranslations), see https://developers.google.com/ml-kit/language/identification/android
+            //If the source language does not exist
+            //and the user did not select auto-detect language, do not allow it.
+            if (!all.contains(from)  && from != AUTO_LANG) {
```
LagradOst: Use useOnlineTranslation, as we should never allow auto on local translation, even if it should not be selectable by UI.  

`!all.contains(from) && !(useOnlineTranslation && from == AUTO_LANG)`

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -2023,13 +2025,17 @@ class ReadActivityViewModel : ViewModel() {
 
             val all = TranslateLanguage.getAllLanguages()
 
-            if (!all.contains(to)) {
+            //If the user wants to translate to a language that doesn't exist,
+            //or wants to auto-detect their own language, do not allow it.
+            if (!all.contains(to) || to == AUTO_LANG) {
```
LagradOst: all does not contain AUTO_LANG, but sure?

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -1941,7 +1941,9 @@ class ReadActivityViewModel : ViewModel() {
         val useOnlineTranslation: Boolean = false
     ) {
         companion object {
+            const val AUTO_LANG = "auto"
             val map = mapOf(
```
LagradOst, line 1945: Maybe have two maps? That way you do not need to `items.removeAt(0)`

```
const val AUTO_LANG = "auto"
val mapOnline = mapOf(AUTO_LANG to "Auto") + map
val listOnline = mapOnline.toList()
fun fromShortToDisplay(from: String): String = mapOnline[from] ?: "Unknown"
```

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -660,19 +659,22 @@ object BookDownloader2Helper {
             try {
                 val page = api.loadHtml(data.url)
 
-                if (!page.isNullOrBlank()) {
+                if (!page.isNullOrBlank() && page.length > 100) {
```
LagradOst: This is an arbitrary limit, please add a comment on why, and how 100 chars is significant.

LagradOst (changes requested):

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
TJMolina: Why should the page contain more than 100 characters? Many websites, when they detect you as a bot, return a message like "F**k off, scraper, I caught you" or something similar. To avoid downloading those chapters, we try to only accept pages with more than 100 characters. Or do you think that's unnecessary?

TJMolina: Anyway, I'll remove it. I don't know, I started thinking that maybe there are novels with extremely short chapters, such as author notes or something similar.

TJMolina:
Vynovel no longer exists; it was completely shut down, so I removed it from the providers.

LagradOst (approved):
Looks good, will merge if nothing is found when testing! ❤️

PR #464, opened 2026-06-05
Improved translations logic

TJMolina:
To make the translation functionality more scalable, I unified everything into a single object that is responsible for deciding which translator to use and how to manage its resources.

### DIFF
--- a/app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
@@ -494,6 +494,10 @@ class ReadActivityViewModel : ViewModel() {
         MutableLiveData<Resource<Boolean>>(null)
     val loadingStatus: LiveData<Resource<Boolean>> = _loadingStatus
 
+    //requires download model observer
+    private val _downloadLanguageRequest = MutableLiveData<MLSettings?>()
+    val downloadLanguageRequest: LiveData<MLSettings?> = _downloadLanguageRequest
+
     private val _chaptersTitles: MutableLiveData<List<UiText>> =
         MutableLiveData<List<UiText>>(null)
     val chaptersTitles: LiveData<List<UiText>> = _chaptersTitles
@@ -1031,7 +1035,7 @@ class ReadActivityViewModel : ViewModel() {
 
     @Throws
     suspend fun requireMLDownload(): Boolean {
-        val settings = mlSettings
+        val settings = MLSettings(from = mlFromLanguage, to = mlToLanguage, mlUseOnlineTransaltion)
         if (settings.isInvalid() || mlUseOnlineTransaltion) return false
         return !TranslationManager.isModelDownloaded(settings.from, settings.to)
     }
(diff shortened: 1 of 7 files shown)


TJMolina:
Oops, I missed a commit

TJMolina:
done

LagradOst (changes requested):
Great that you split up the logic into a separate file!

However, this code has an issue with the structure of TranslationManager. TranslationManager can not be a singleton (aka `object TranslationManager` instead of `class TranslationManager`), as that fucks up all translations if you launch two instances of the app at the same time. 

You must make TranslationManager into a class that is used as a regular object in ReadActivityViewModel like the old mlTranslator. That way many instances of the app can be launched without messing up the translation, as they will not share the translation state.

app/src/main/java/com/lagradost/quicknovel/util/Translation/TranslationManager.kt
```diff
@@ -0,0 +1,102 @@
...
+        val client = translator ?: prepareModel(from, to) ?: throw Exception("Model not available")
+
+        // Safe paragraph processing loop to prevent single failures from crashing whole chapters
+        return textList.mapIndexed { index, text ->
+            if (!text.trim().any{it.isLetter()}) return@mapIndexed text
+            try {
+                progress(index, textList.size)
+                Tasks.await(client.translate(text))
+            } catch (e: Exception) {
+                // Return original untranslated paragraph as fallback item on error
```
LagradOst: I do not really like this behavior, as it makes it very confusing.

app/src/main/java/com/lagradost/quicknovel/util/Translation/TranslationManager.kt
```diff
@@ -0,0 +1,102 @@
...
+    }
+    suspend fun prepareModel(from: String, to: String): Translator? {
+        try {
+            // Close previous translator if language pair changed
+            if (currentFrom != from || currentTo != to) {
+                translator?.close()
+            }
+
+            val sourceTag = TranslateLanguage.fromLanguageTag(from) ?: TranslateLanguage.ENGLISH
+            val targetTag = TranslateLanguage.fromLanguageTag(to) ?: TranslateLanguage.SPANISH
```
LagradOst: The default here is weird, it should throw an error.

app/src/main/java/com/lagradost/quicknovel/ReadActivityViewModel.kt
```diff
@@ -494,6 +494,10 @@ class ReadActivityViewModel : ViewModel() {
         MutableLiveData<Resource<Boolean>>(null)
     val loadingStatus: LiveData<Resource<Boolean>> = _loadingStatus
 
+    //requires download model observer
+    private val _downloadLanguageRequest = MutableLiveData<MLSettings?>()
+    val downloadLanguageRequest: LiveData<MLSettings?> = _downloadLanguageRequest
```
LagradOst: Unused?

LagradOst (commented):
Unsure if you are finished, but you need to change `object TranslationManager` into `class TranslationManager`

TJMolina:
Damn, I always forget something haha. I'll do it today.

TJMolina:
I'm still working on this PR. I'll be adding more changes

TJMolina:
I'll have to redo this PR when the ReadActivity is also Compose

PR #477, opened 2026-07-02
Providers and Epubs Fixes

TJMolina:
I made some general fixes to some providers (I'll keep updating this PR since there's still more to do).

For now, there are two main highlights:

1. I fixed out-of-memory issues when reading very large EPUBs.
2. I had to do something pretty crazy for Anna's Archive. They fixed the issue where their EPUB URLs were easily accessible, so the only option left was to work around their anti-bot protections. They're either really good at what they do, or I'm really bad haha. I tried updating the Cloudflare Killer and WebView logic, but no matter how much I researched, I couldn't get it working. As a last resort, I implemented a new approach that uses WebView directly for Anna's Archive. It works extremely well, and it could eventually be used for other equally difficult providers like JNovels, but that's something for the future. The important thing is that Anna's Archive keeps working, because it's simply too good of a provider to lose.

### DIFF
--- a/app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
@@ -1925,6 +1925,7 @@ object BookDownloader2 {
         count: Int
     ): Boolean? {
         val originalBitmap = imageXObject.image ?: return null
+        if (originalBitmap.width <= 0 || originalBitmap.height <= 0) return null
         if (!originalBitmap.isRecycled) {
             try {
                 val maxRes = 1200
--- a/app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
@@ -50,6 +50,7 @@ import com.lagradost.quicknovel.BookDownloader2Helper.IMPORT_SOURCE_PDF
 import com.lagradost.quicknovel.BookDownloader2Helper.createQuickStream
 import com.lagradost.quicknovel.BookDownloader2Helper.generateId
 import com.lagradost.quicknovel.BookDownloader2Helper.getDirectory
+import com.lagradost.quicknovel.BookDownloader2Helper.getSafeByteArray
 import com.lagradost.quicknovel.CommonActivity.activity
 import com.lagradost.quicknovel.CommonActivity.showToast
 import com.lagradost.quicknovel.DataStore.getSharedPrefs
@@ -93,7 +94,6 @@ import kotlinx.coroutines.sync.withPermit
 import kotlinx.coroutines.withContext
 import me.ag2s.epublib.domain.Author
 import me.ag2s.epublib.domain.EpubBook
-import me.ag2s.epublib.domain.MediaType
 import me.ag2s.epublib.domain.MediaTypes
 import me.ag2s.epublib.domain.Resource
 import me.ag2s.epublib.epub.EpubReader
@@ -105,6 +105,7 @@ import java.io.FileOutputStream
 import java.io.IOException
 import java.util.concurrent.ConcurrentHashMap
 import kotlin.coroutines.cancellation.CancellationException
+import kotlin.io.readBytes
 import kotlin.time.Duration.Companion.milliseconds
 
 enum class DownloadActionType {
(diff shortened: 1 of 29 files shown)


TJMolina:
Added related novels and reviews to all compatible English and Spanish providers

TJMolina:
I'll make this branch compatible with master once the Compose changes are finished, so I can avoid doing all that extra work 😅

LagradOst:
Holy this is a large provider update 🔥

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -271,21 +346,24 @@ object BookDownloader2Helper {
...
-            val cr = activity.contentResolver ?: return false
-            val fileUri =
-                cr.getExistingDownloadUriOrNullQ(relativePath, displayName) ?: return false
-            val fileLength = cr.getFileLength(fileUri) ?: return false
-            return fileLength != 0L
-        } else {
-            val normalPath =
-                "${Environment.getExternalStorageDirectory()}${fs}$relativePath$displayName"
+        return try {
+            val (baseDir, _) = activity.getBasePath()
```
LagradOst: Use this instead
```kotlin
fun hasEpub(activity: Activity?, name: String): Boolean {
        if (activity == null) return false

        if (!activity.checkWrite()) {
            activity.requestRW()
            return false
        }

        try {
            val subDir =
                activity.getBasePath().first ?: getDefaultDir(activity) ?: throw IOException("No file")
            val displayName = "${sanitizeFilename(name)}.epub"
            val foundFile = subDir.findFileOrThrow(displayName)

            return foundFile.uri() != null
        } catch (_ : Throwable) {
            return false
        }
    }
```

app/src/main/java/com/lagradost/quicknovel/providers/WuxiaClickProvider.kt
```diff
@@ -309,9 +315,54 @@ class WuxiaClickProvider :  MainAPI() {
...
                 it.text().trim().takeIf { text ->  !text.isEmpty() }
             }
+            related = getRelated()
         }
     }
 
+    suspend fun getRelated(): List<SearchResponse> {
+        val url = "$secondUrl/api/novels/$novelId/recommendations/"
+        val response = app.get(url).parsed<RelatedResponse>()
+        println(response)
```
LagradOst: Remove

app/src/main/java/com/lagradost/quicknovel/providers/FenrirRealProvider.kt
```diff
@@ -131,10 +153,31 @@ class FenrirRealProvider:  MainAPI() {
...
         }
     }
 
 
+    override suspend fun loadReviews(
+        url: String,
+        page: Int,
+        showSpoilers: Boolean
+    ): List<UserReview> {
+        if (libraryId.isEmpty()) return emptyList()
```
LagradOst, line 166: This is really bad, you should not have any state here!

app/src/main/java/com/lagradost/quicknovel/providers/FenrirRealProvider.kt
```diff
@@ -20,6 +21,9 @@ class FenrirRealProvider:  MainAPI() {
     override val hasMainPage = true
     //I don't know why, but this fixes the timeout issue for this provider
     override val rateLimitTime = 3000L
+    override val hasReviews = true
+    override val usesCloudFlareKiller = true
+    var libraryId = ""
```
LagradOst, line 26: If you really really need to store some state, then put it in a ConcurrentHashMap and map the url to the content needed. Because the user can have several pages open on the same time.

app/src/main/java/com/lagradost/quicknovel/providers/LibReadProvider.kt
```diff
@@ -17,6 +30,8 @@ open class LibReadProvider : MainAPI() {
     override val iconId = R.drawable.icon_libread
 
     override val iconBackgroundId = R.color.libread_header_color
+    override val hasReviews = true
+    var novelId = ""
```
LagradOst: See the comment above.

app/src/main/java/com/lagradost/quicknovel/providers/LightNovelWorldProvider.kt
```diff
@@ -8,26 +8,95 @@ import com.lagradost.quicknovel.LoadResponse
...
 
 class LightNovelWorldProvider : MainAPI() {
     override val name = "LightNovelWorld"
     override val mainUrl = "https://lightnovelworld.org"
     override val hasMainPage = true
     override val iconId = R.drawable.icon_lightnovelworld
     override val iconBackgroundId = R.color.colorPrimaryWhite
-
+    override val hasReviews = true
+    var novelId = ""
```
LagradOst: ^

app/src/main/java/com/lagradost/quicknovel/providers/NovelBuddyProvider.kt
```diff
@@ -6,26 +6,116 @@ import com.lagradost.quicknovel.LoadResponse
...
-    private val apiUrl = "https://api.novelbuddy.com/titles"
+    override val mainUrl = "https://novelbuddy.me"
+    private val apiUrl = "https://api.novelbuddy.me/titles"
     override val iconId = R.drawable.icon_novelbuddy
     override val iconBackgroundId = R.color.novelBuddyColor
     override val lang = "en"
     override val hasMainPage = true
-
+    override val hasReviews = true
+    var novelId = ""
```
LagradOst: ^

app/src/main/java/com/lagradost/quicknovel/providers/NovelFireProvider.kt
```diff
@@ -22,7 +25,9 @@ class NovelFireProvider:  MainAPI() {
     override val iconId = R.drawable.icon_novelfire
     override val rateLimitTime = 500L
     override val hasMainPage = true
-
+    override val hasReviews = true
+    var novelId: String = ""
+    var nextPosts: String = ""
```
LagradOst: ^

app/src/main/java/com/lagradost/quicknovel/providers/NovelLightProvider.kt
```diff
@@ -26,6 +28,8 @@ class NovelLightProvider:  MainAPI() {
     override val iconId = R.drawable.icon_novelight
     override val iconBackgroundId = R.color.novelightColor
     override val hasMainPage = true
+    override val hasReviews = true
+    var novelId = ""
```
LagradOst: ^

app/src/main/java/com/lagradost/quicknovel/providers/PawReadProver.kt
```diff
@@ -1,25 +1,29 @@
...
 
 class PawReadProver : MainAPI() {
     override val name = "PawRead"
     override val mainUrl = "https://pawread.com"
     override val iconId = R.drawable.pawread
-
     override val hasMainPage = true
-
+    override val hasReviews = true
+    var novelId = ""
```
LagradOst: ^

app/src/main/java/com/lagradost/quicknovel/providers/RanobesProvider.kt
```diff
@@ -17,6 +15,8 @@ class RanobesProvider : MainAPI() {
     override val iconBackgroundId = R.color.white
     override val usesCloudFlareKiller = true
     override val rateLimitTime = 500L
+    override val hasReviews = true
+    var novelId = ""
```
LagradOst: ^

app/src/main/java/com/lagradost/quicknovel/providers/WattpadProvider.kt
```diff
@@ -16,7 +18,7 @@ class WattpadProvider : MainAPI() {
     override val iconId = R.drawable.icon_wattpad
     override val iconBackgroundId = R.color.wattpadColor
     override val rateLimitTime = 500L
-
+    var novelId = ""
```
LagradOst: ^

app/src/main/java/com/lagradost/quicknovel/providers/WtrLabProvider.kt
```diff
@@ -27,9 +31,9 @@ class  WtrLabProvider : MainAPI() {
     override val lang = "en"
     override val iconId = R.drawable.icon_wtrlab
     override val hasMainPage = true
-    override val hasReviews = false
     override val usesCloudFlareKiller = true
-
+    override val hasReviews = true
+    var novelId = ""
```
LagradOst: ^

app/src/main/java/com/lagradost/quicknovel/providers/WuxiaClickProvider.kt
```diff
@@ -1,25 +1,31 @@
...
 import org.jsoup.nodes.Element
 
 class WuxiaClickProvider :  MainAPI() {
     override val name = "WuxiaClick"
     override val mainUrl = "https://wuxia.click"
+    val secondUrl = "https://wuxiaworld.eu"
     override val iconId = R.drawable.icon_wuxiaclick
     override val iconBackgroundId = R.color.wuxiacliColor
+    override val hasReviews = true
+    var novelId = ""
```
LagradOst: ^

app/src/main/java/com/lagradost/quicknovel/ui/settings/SettingsFragment.kt
```diff
@@ -110,6 +110,7 @@ class SettingsFragment : Fragment(), SearchableSettings by SettingScreen() {
             return when {
                 path.isNullOrBlank() -> getDefaultDir(context)
                 path.startsWith("content://") -> SafeFile.fromUri(context, path.toUri())
+                path.startsWith("/storage") -> SafeFile.fromFilePath(context, path)
```
LagradOst: Was this a problem?

app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
```diff
@@ -1850,13 +1916,26 @@ object BookDownloader2 {
...
-        if (!bitmap.isRecycled) {
+        val originalBitmap = imageXObject.image ?: return null
+        if (!originalBitmap.isRecycled) {
             try {
-                val outStream = ByteArrayOutputStream()
+                val maxRes = 1200
+                val shouldScale = originalBitmap.width > maxRes || originalBitmap.height > maxRes
+
+                val finalBitmap = if (shouldScale) {
+                    val ratio = originalBitmap.width.toFloat() / originalBitmap.height.toFloat()
```
LagradOst, line 1935: Please do a sanity check before maybe dividing by zero.

LagradOst (changes requested):
Please make the providers stateless, or at least use a concurrenthashmap to store the state so many can be open at the same time 👍 

Also given that many providers use the same-ish code you might want to make that into a helper class. I realize that hasEpub was broken and it was good that you found that, however try to use the code I wrote instead as it mirrors the createEpub code.

TJMolina:
done

app/src/main/java/com/lagradost/quicknovel/providers/FenrirRealProvider.kt
LagradOst, line 166: This one is still active?

LagradOst: I will just fix it rq

TJMolina: Damn, I always end up missing something. Sorry!

LagradOst: No problem ❤️  I fixed this and pushed an update!

Issue #481, opened 2026-07-17
Download problems (0/0) showing


DeepProxi07:
<img width="1080" height="2408" alt="Image" src="https://github.com/user-attachments/assets/50e50674-8205-4666-a7b8-1bc90f4bf286" />

JoeJoeflyn:
Thanks for the report. Could you share a few more details so I can reproduce this?

  1. Does the "0/0" stay forever, or does it eventually change to a real number (like 5/120)?
  2. App version — can you check in Settings or the About screen?
  3. Did the download actually start (any progress bar / loading indicator), or is it just sitting in the list doing nothing?

The "0/0" usually means the downloader hasn't fetched the chapter list yet, but I want to confirm whether it's stuck there or just briefly showing before the real count loads.

DeepProxi07:
App version is letest. The chapter were download before when I tried to update them it went 0/0 . I search for the same novel from other sources there They were fine. I think It might've been a problem with source 'FreeWebNovel' . Only novels downloaded form that source are effected but others are fine.

TJMolina:
I've already submitted a pull request to fix this issue, so don't worry.

https://github.com/LagradOst/QuickNovel/pull/477

DeepProxi07:
Thank you

PR #484, opened 2026-07-18
Fix 0/0 download progress on certain FreeWebNovel novels

JoeJoeflyn:
## Summary
- Restores DOM-based chapter extraction as a fallback for FreeWebNovel pages that lack `window.chapterPagination`
- Fixes issue #481 where certain novels showed permanent "0/0" download progress
- Pages with `window.chapterPagination` are unaffected

## Root cause
PR #459 changed chapter list extraction from DOM-based scraping to script-based parsing (`window.chapterPagination`). However, many FreeWebNovel novel pages don't contain this script — they list all chapters directly in the DOM under `div.m-newest2 ul.ul-list5 li`. For those pages, `getChapterList()` returned 0 chapters, which overwrote `DOWNLOAD_TOTAL` with 0.

## Fix
Restores the pre-#459 DOM-based extraction as a fallback when `window.chapterPagination` is absent.
screenshot-2026-07-18_20-00-21

### DIFF
--- a/app/src/main/java/com/lagradost/quicknovel/providers/LibReadProvider.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/providers/LibReadProvider.kt
@@ -9,7 +9,7 @@ open class LibReadProvider : MainAPI() {
     override val mainUrl = "https://libread.com"
 
     //for some reason, now is freewebnovel
-    val secondUrl = "https://freewebnovel.com"
+    val secondUrl = "http://10.0.2.2:8080"
     override val hasMainPage = true
 
     open val removeHtml = false // because the two sites use .html or not for no reason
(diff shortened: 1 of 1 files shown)


TJMolina:
Hi friend, the url http://10.0.2.2:8080 does not exist, also this error was already fixed in my PR Providers and Epubs Fixes don't worry

JoeJoeflyn:
My bad, that URL was only used for local testing on the emulator and was reverted before committing — it's not in the actual PR. The only change here is the DOM fallback in LibReadProvider.kt. I'll take a closer look at your PR to avoid duplicating work.

If the issue is already solved by your PR, could you comment on issue #481 or link it so other maintainers know it's handled? Thanks!

TJMolina:
Sure, I'll do that. Thanks for the collaboration as well, Cheers. If you want to see where it's fixed in the PR, check the getChapters function in LibReadProvider

JoeJoeflyn:
Thanks! I'll check the getChapters function in LibReadProvider to see your approach. Happy to contribute more to the project. Cheers!

Issue #462, opened 2026-05-30
click to copy the source name


rockyxwall:
when you click on the title in the novel page it should get copied. more complex, when clicking on the title it searches for the title on every site. similar to mihon

JoeJoeflyn:
i'll take a look for this

PR #486, opened 2026-07-18
Add click-to-copy and long-press-to-search on novel titles

JoeJoeflyn:
## Summary
- **Click novel title** → copies the title to clipboard with a "copied!" toast
- **Long-press novel title** → navigates to the search page, auto-fills the search bar with the title, and searches across all enabled providers

Related issue: #462

### Changes
| File | Change |
|------|--------|
| `ResultFragment.kt` | Added `setOnClickListener` (copy) and `setOnLongClickListener` (search) on `resultTitle` |
| `SearchFragment.kt` | Added `SearchFragmentExtras` static variable to pass query between fragments |
| `SearchScreen.kt` | Added `pendingQuery` parameter + `LaunchedEffect` to auto-fill search bar and trigger search |
| `fragment_result.xml` | Moved `result_info_header` after `NestedScrollView` so title receives touches (was intercepted by ScrollView z-order). Added `fitsSystemWindows` to keep title below status bar. Added `longClickable="true"` on title. |

### Why the layout change?
In a `FrameLayout`, later children draw on top. The `NestedScrollView` was the second child, covering the header and swallowing all touch events. Moving the header after the ScrollView ensures it receives clicks and long-presses.

https://github.com/user-attachments/assets/c4f01fce-bce8-49e5-9da0-3d3eb8f10708

### DIFF
--- a/app/src/main/java/com/lagradost/quicknovel/ui/result/ResultFragment.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/ui/result/ResultFragment.kt
@@ -40,6 +40,7 @@ import com.lagradost.quicknovel.ui.ReadType
 import com.lagradost.quicknovel.ui.SortingMethodAdapter
 import com.lagradost.quicknovel.ui.mainpage.MainAdapter
 import com.lagradost.quicknovel.ui.mainpage.MainPageFragment
+import com.lagradost.quicknovel.ui.search.SearchFragmentExtras
 import com.lagradost.quicknovel.ui.setRecycledViewPool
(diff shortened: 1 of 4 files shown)


TJMolina:
I think it's a very useful feature, but wouldn't it make more sense the other way around? I mean, we usually copy text by long-pressing it, don't we? I think tapping to copy would feel a bit unusual (just my personal opinion haha)

JoeJoeflyn:
@TJMolina Yep, the copy one is redundant and doesn't add any value. We could just have auto-search with one tap instead, that's what I already added. I can remove the redundant changes, we only need the search one

PR #490, opened 2026-07-20
New providers

TJMolina:
Added **NovelPhoenix**, **Ncode**, and **LuxoScan** (I may add more later).

**Notes:**

1. I removed a few providers because they were only used as base classes for other providers. Instead, I merged their code into one of the providers that is actually used and made that provider an `open` class.

2. **Ncode** doesn't provide cover images, but it's still a very valuable provider because almost every Japanese web novel originates there. I'm also working on adding Gemini-powered translation, so having this provider will be even more beneficial in the future.

3. There are several providers that are no longer used because they no longer work, and there are also some providers that are still included but are broken (such as **Novels Online**). Would you be okay with me removing them to keep the app cleaner?


LagradOst:
Crazy work! I will try to merge this and #477 before next update for a huge UI + provider update!

app/src/main/java/com/lagradost/quicknovel/providers/MeioNovelProvider.kt
```diff
@@ -118,121 +95,81 @@ open class MeioNovelProvider : MainAPI() {
...
         val document = app.get(url).document
-        val name =
-            document.selectFirst("div.post-title > h1")?.text()?.replace("  ", " ")
-                ?.replace("\n", "")
-                ?.replace("\t", "") ?: return null
-        val data = getChapters(url)
-        return newStreamResponse(url = url, name = name, data = data) {
+        val name = document.selectFirst("div.post-title > h1")?.text()?.clean() ?: return null
+
+        return newStreamResponse(url, name, getChapters(url)) {
```
LagradOst: Wrong order on url and name.

app/src/main/java/com/lagradost/quicknovel/providers/MeioNovelProvider.kt
```diff
@@ -84,32 +67,26 @@ open class MeioNovelProvider : MainAPI() {
...
             else -> "novel-genre/$tag"
         }
-        val url =
-            "$mainUrl/$order/page/$page/${if (orderBy == null || orderBy == "") "" else "?m_orderby=$orderBy"}"
-
+        val url = "$mainUrl/$order/page/$page/${if (orderBy.isNullOrBlank()) "" else "?m_orderby=$orderBy"}"
         val document = app.get(url).document
-
-        //""div.page-content-listing > div.page-listing-item > div > div > div.page-item-detail"
+        println(document)
```
LagradOst: Remove

LagradOst (changes requested):

app/src/main/java/com/lagradost/quicknovel/providers/NovLoveProvider.kt
```diff
@@ -1,9 +1,48 @@
...
+
+    override suspend fun load(url: String): LoadResponse? {
+        //This is for migrating NovLove URLs.
+        val path = url.substringAfter("://").substringAfter("/", "")
+        val adaptedPath = path.replaceFirst("novel/", "read/")
+        val finalUrl = "${mainUrl}/$adaptedPath".removeSuffix("/")
+        val document = app.get(finalUrl).document
+        val name = document.selectFirst("div.post-title > h1")?.text()?.clean() ?: return null
+
+        return newStreamResponse(url, name, getChapters(url)) {
```
LagradOst: Same error?

LagradOst (changes requested):

LagradOst (approved):

PR #498, opened 2026-07-25
Providersfix 3

TJMolina:


LagradOst (approved):
Looks good, not tested yet.

TJMolina:
I forgot to comment the code, I'll do it tomorrow. But basically it seems like there are versions of Cloudflare or something like that are more meticulous. I had to modify WebViewResolver and CloudflareKiller to collect the headers and reformulate the request with all the data properly ordered and added so that okhttp would work

I also added a debounce because sometimes a page gets the "correct" token and then reloads the page and changes it to another one, which is the real one but it never gets obtained

TJMolina:
Better to delete this PR and make another one correctly. This is because ScribbleHub and wtrlab require carefully checking the cloudflare killer and the webview resolver, so I'll make a separate PR.

PR #502, opened 2026-07-28
Fix background downloads not working

jpz-u:
Downloads stop shortly after the app leaves the foreground, and the notification
freezes or disappears with no error. There are three independent causes.

### 1. The Compose download button never used WorkManager

`ResultViewModel2` and `DownloadViewModel2` called `BookDownloader2.downloadWorkThread(...)`
directly inside `viewModelScope.launch { }`. That scope is cancelled when the ViewModel is
cleared, so the download dies as soon as the screen goes away — navigating back is enough,
the app does not even need to be backgrounded.

`DownloadFileWorkManager.download()` already existed, but only the older
`ResultViewModel` / `refreshCard` paths reached it.

Both call sites now enqueue work instead, via a new `ImmutableSearchResponse` overload.

### 2. The worker was never a foreground worker

`DownloadFileWorkManager` had no `getForegroundInfo()` and never called `setForeground()`.
A plain `CoroutineWorker` gets roughly ten minutes before WorkManager stops it, and once the
process is cached the network is throttled. The class comment says it exists to solve exactly
this problem, but the mechanism was never implemented — `DownloadNotificationService` is
declared `foregroundServiceType="dataSync"` in the manifest yet only ever handled the
pause/resume/stop notification taps.

The worker now promotes itself for `ID_DOWNLOAD` and `ID_REFRESH_DOWNLOADS`:

- `getForegroundInfo()` returns the three argument `ForegroundInfo` with
  `FOREGROUND_SERVICE_TYPE_DATA_SYNC` on API 29+, the two argument one below that
  (`minSdkVersion` is 23).
- The notification channel is created eagerly, before the first `setForeground()`, otherwise
  the foreground notification is invalid.
- `setForeground()` is wrapped in `try`/`catch`. It throws
  `ForegroundServiceStartNotAllowedException` when the app is not allowed to start a
  foreground service, and an uncaught throw would kill the worker outright.

`AndroidManifest.xml` also merges `foregroundServiceType="dataSync"` onto
`androidx.work.impl.foreground.SystemForegroundService`. WorkManager does not declare a type
itself, and Android 14+ rejects `startForeground` with a type that is not in the manifest.
Without this the promotion throws and is silently swallowed by the catch above.

### 3. The download path depended on a `WeakReference<Activity>`

`downloadWorkThread(StreamResponse, …)` opened with `val context = activity ?: return`, and
the `EpubResponse` and `Uri` overloads did the same with `activity?.filesDir ?: return`.
Once past that line the local reference is strong, so a download already in flight kept
running — but a download that *started* while `CommonActivity.activity` was already null
aborted instantly and silently.

Worse, `createNotification` re-read the global on every single call, so progress notifications
died mid download while the file writes carried on. That asymmetry is why the symptom looks
like "the download stopped" when it had not.

`Context` is now threaded explicitly through the `downloadWorkThread` overloads,
`createNotification`, `copyAllData` and `deleteNovel`, following the pattern
`downloadPDFWorkThread` already used. `filesDir` is identical for an Activity and the
Application context, so this is behaviour preserving.

The remaining `activity` uses are `turnToEpub` / `openEpub` / `hasEpub` / `createQuickStream`,
which are user initiated and correctly Activity scoped.

### Also

- The in memory `workData` map does not survive process death, so `popWork` returned null and
  the worker gave up with `Result.failure()`. The download id is now stored in the worker
  `Data`, and on a miss the request is rebuilt from the persisted `DOWNLOAD_FOLDER` key.
  Capped at three attempts so a dead source cannot retry forever. This also covers the
  "worker stopped mid download and got rescheduled" case.
- The unique work name includes the download id. `refreshDownloads()` enqueues in a loop, and
  with a purely millisecond based name a batch collides and chains under `APPEND`, where one
  failure cancels every download behind it.

### Testing

Builds clean against `compileSdk 37` / AGP 9.2.1. The merged manifest was checked to confirm
`SystemForegroundService` carries `foregroundServiceType="dataSync"` and stays enabled.

Verified on a Samsung Galaxy S25 Ultra (Android 15 or later, One UI): downloads started from
the app now continue while it is in the background, with the ongoing notification present and
progressing. That covers the Android 14+ path where `startForeground` requires the service
type to be declared in the manifest, on a vendor build with aggressive background management.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -48,6 +48,12 @@
             android:exported="false"
             android:foregroundServiceType="dataSync" />
 
+        <!-- Android 14+ rejects startForeground with a type that is not declared here -->
+        <service
+            android:name="androidx.work.impl.foreground.SystemForegroundService"
+            android:foregroundServiceType="dataSync"
+            tools:node="merge" />
+
         <receiver
             android:name="androidx.media.session.MediaButtonReceiver"
             android:exported="true">
--- a/app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
+++ b/app/src/main/java/com/lagradost/quicknovel/BookDownloader2.kt
@@ -361,7 +361,7 @@ object BookDownloader2Helper {
     }
 
     fun copyAllData(
-        activity: Activity?,
+        context: Context?,
         fromAuthor: String?,
         fromName: String,
         fromApiName: String,
@@ -369,7 +369,7 @@ object BookDownloader2Helper {
         toName: String,
         toApiName: String
     ) {
-        if (activity == null) return
+        if (context == null) return
         val sFromApiName = sanitizeFilename(fromApiName)
         val sFromAuthor = if (fromAuthor == null) "" else sanitizeFilename(fromAuthor)
         val sFromName = sanitizeFilename(fromName)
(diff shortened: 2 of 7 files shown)


LagradOst:
I always had this planned on this refactor, but never got around to do it. While this looks AI generated, it works 👍

PR #504, opened 2026-07-29
CloudflareKiller changes

TJMolina:
This PR introduces several improvements and modifications to **CloudflareKiller** and **WebViewResolver** to make them more robust when dealing with Cloudflare challenges.

#### 1. ScribbleHub

ScribbleHub would not work no matter what I tried, so I investigated the issue with the help of AI. The solution was to clone the entire WebView request as faithfully as possible, including the exact order of headers and cookies.

To achieve this, I added a map to store the headers captured from the WebView and modified `CloudflareKiller.proceed()` so it reconstructs the entire request before sending it.

The `request.headers.forEach` line is currently commented out because I tested the implementation without filtering custom headers and everything worked correctly. I'm not sure whether filtering custom headers will be necessary in the future, so I decided to leave that code commented out for now.

#### 2. WTRLab
Its chapter content requires two requests:

* One request fetches the container page that contains the chapter ID.
* The second request is a POST request that retrieves the actual chapter.

The problem is that this POST request is protected by Cloudflare in a different way. Instead of always returning JSON, it can return either JSON or HTML. After roughly 15 requests, the server starts returning an HTML page containing a Cloudflare iframe challenge.

Because of this, `CloudflareKiller` fails to detect the challenge. `looksLikeCloudflareChallenge()` intentionally does not read the entire response body, and even after increasing the amount of the document it reads, I still couldn't get it to solve the challenge for this provider.

To solve this, I modified `WebViewResolver` to support an alternative Cloudflare bypass mode. It now accepts an optional script (a new parameter). When a script is provided, the WebView remains on the page until the desired content appears.

In WTRLab's case, the WebView waits until the chapter content becomes available and then returns its HTML. Because of this change, the return type changed from:

`Pair<Request?, List<Request>>`

to:

`Pair<Any?, List<Request>>`

An additional benefit is that the `cf_clearance` cookie is obtained together with the chapter HTML. Once that happens, the next ~15 POST requests can be performed normally through `app.post()` without having to invoke `WebViewResolver` again.

#### 3. Cookie utilities

I moved the cookie-related utilities out of `CloudflareKiller` because they are now used outside of that class, and `CloudflareKiller` itself was becoming too large.

I also created a new `network/utils` package. As QuickNovel continues to grow, the current `utils` package is becoming increasingly crowded. Since `CookiesUtils` is specifically related to networking, I thought it made more sense to organize it under a dedicated network utilities package.


TJMolina:
I forgot to mention that `CloudflareKiller` also failed in WTRLab's case because it always returned a **200 OK** status code, so it was never recognized as a failed request.
I also tried modifying the `looksLikeCloudflareChallenge` logic so it would always accept a response body that looked like a Cloudflare challenge, but that didn't work either. I also don't think it's a good idea, because if, for example, a page's title happened to be **"One Moment"**, it would be incorrectly identified as a Cloudflare challenge even when it wasn't.

TJMolina:
Damn, I forgot the ScribbleHub icon

TJMolina:
Lnori implemented the decoy cookie logic, and Cloudflare had a bug that was causing all the tests to fail. It's already been fixed

LagradOst (approved):
This looks good and works well when tested. I have put this pr on the back-burner simply due to the fact that CloudflareKiller+WebViewResolver is a more sensitive area of code compared to a provider.

PR #506, opened 2026-08-03
Epub improved

TJMolina:
When reading an EPUB created by QuickNovel in any old external reading app, paragraph spacing didn't work.
This PR fixes that by adding CSS styles to the EPUB files. These styles are recognized by external readers and correctly apply paragraph spacing.


TJMolina:
sorry, i deleted this by accident

LagradOst (approved):
Looks good, will have to test it before merge.

PR #514, opened 2026-08-15
Fix TTS stopping in background on Android 14+

faadiallop:
This pull request fixes an issue where the TTS stops playing when the screen is turned off or the app is switched to the background on Android 14 devices.

### Changes:
- Activated the `MediaSession` inside `TTSNotifications.kt` to comply with the Android 14 requirement that `mediaPlayback` foreground services must hold an active media session.
- Added the `FOREGROUND_SERVICE_TYPE_MEDIA_PLAYBACK` explicitly when calling `startForeground` in `TTSNotificationService.kt` for devices running Android 10+.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,33 @@
+name: Build Android APK
+
+on:
+  push:
+    branches:
+      - fix-tts-android-14
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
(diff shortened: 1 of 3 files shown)


LagradOst (changes requested):
From my testing, this does not change any behavior at all. Are you sure this makes TTS better for you, have you tested yourself on android 14+?

faadiallop:
Yeah, doesn't seem to fix the issue. Closing until it's fully debugged.

TJMolina:
Hi, what you're reporting is fixed by turning off battery optimization for QuickNovel

PR #519, opened 2026-08-18
More tags

TJMolina:
Uh, let me explain myself before you dismiss this PR, please! I know it looks really long, but it's actually not that big. It's just that I included the removal of a lot of unused imports in the PR, and I also changed the structure of several screens a bit, for example, how dialogs are displayed. There's an explanation for that further down.

Please don't think this is just "a lot of text and AI garbage." This message is 100% written by me (just translated by AI), and the code is 90% mine. If there's garbage, it's human garbage haha, but there isn't any garbage! (personal opinion).

Anyway, here's my explanation:

This PR does several related things. One thing couldn't be done without doing another.

The main goal of this PR is to integrate the ability to customize your bookmarks. It allows you to create, rename, move, etc.

The problem is that implementing this required `MainActivity` to support Compose, and the novel preview also had to be Compose. So this PR makes `MainActivity` hybrid until there's a plan to fully migrate it to Compose. In the meantime, this PR already does most of that migration, but not completely.

Another problem was that bookmark editing was another sub-dialog inside the novel preview dialog, so that also had to be completely migrated to Compose.

I tried to respect the state management you've already implemented in the parts that were already migrated to Compose.

One problem that came from this was that when a user wanted to create a bookmark, they had to wait for the preview to load. So I had to use the cache, which already contained previous information about the novel, to provide a quick response and show the preview faster. The data will still be updated in the background anyway.

However, this caused the inconvenience that the status wasn't reflected immediately, so variables like `ResultCached` and `ImmutableState` also had to be modified to include the novel's state in the cache.

Another problem with Compose is that making a view draggable is difficult without a library, so I partially dropped that functionality. Now there are only buttons, which I think only the more curious users will notice. These buttons allow you to move the bookmarks around and change their position however you want.

Another problem is that I have absolutely no idea why the descriptions in the preview don't respect tags like `<strong>`, even though I used the `String.html()` extension just like in the result screen.

Regarding the result screen, you're still working on it, so I didn't add the bookmark logic there. However, if you're interested in this implementation, I would highly recommend integrating this PR first and then continuing with the result screen. Otherwise, we'd have to do the work twice for no reason, since it would have to be removed anyway.

I shouldn't forget to mention that while migrating `MainActivity` to Compose, I had to create a `MainViewModel` to handle the global state and functions for the preview and bookmarks.

Oh, and I said I was going to mention this, sorry. For the dialogs, I didn't want to create a bunch of variables that individually decided whether each one should be shown or not. Instead, I made the dialogs that are shown one at a time part of a sealed class, so an exhaustive `when` determines which dialog should be displayed. I thought this was much easier than changing individual variables.


TJMolina:
I forgot to mention that it also includes some leftover XML from the previous version of this PR. I'll remove it.

LagradOst (commented):
Good migration to compose, I see that you have invested a lot of work into making this work. Very funny pr comment. 

However, I have some high level concerns that should make this pr a lot smaller, and more maintainable. 

I am not a huge fan of MainViewModel, given that ResultViewModel2 should instead support using a partial ImmutableSearchResponse. Impo MainViewModel should be removed with better support for ResultViewModel2. As a consequence ResultState instead of NovelPreviewData ect

BottomPreview should also not be a "common" view as it is a specific dialog for a single use-case that should interface directly with a ResultViewModel2. Having several different callbacks instead of a ResultViewModel2 onaction callback is overengineering impo for a more flexable design that will never be used. In general I think it is good to expose `action: (XAction) -> Unit,` instead of a `onClick: () -> Unit,` for a specific view.

TJMolina:
Okayyy, I'll completely remove `MainViewModel` and move everything to `ResultViewModel2`. But I have two questions:

Should the preview still be shown by `MainActivity`? Because it's the only place where I can show the same dialog from all three places where it's used (`DownloadScreen`, history, and result). If so, Is it okay if I create a new instance of `ResultViewModel2`, or should I do something like a shared ViewModel?

My other question is that `ImmutableSearchResponse` initially needs to be loaded using the URL, which removes the quick response I was getting from `ResultCached`. Is it okay if I implement that logic in `ResultViewModel2` as well? I mean, allowing it to receive different data models, such as `ResultCached`, to provide a quick response for the preview.

Now that I think about it, regarding the preview and bookmark dialog, would it be okay if I created a package under `ui` called `bookmark` containing something like `BookmarkScreen`, but internally it would contain the two dialogs?

TJMolina:
Done. I removed MainViewModel and moved all the logic (except the UI) to ResultViewModel. I made MainActivity share the ViewModel with ResultScreen so it can control the dialogs' visibility states. However, I only implemented this for ResultScreen, not for the old view.

I also created a preview package under ui > result where I keep the Compose code for the preview and bookmark dialogs.

TJMolina:
Why do I use multiple versions of loadResult? It's to give the user a fast response!

Imagine a user wants to move several novels between different bookmarks. They would have to wait for the online data to load every single time first. That's way too annoying, so I made it accept data from other models such as SearchResponse, ResultCached, etc., so the preview can be shown immediately. The data is still refreshed in the background afterward.

TJMolina:
I modified `ResultViewModel` to only use `ImmutableSearchResponse`, making `MainActivity` use `ImmutableSearchResponse.from()` to convert the data into an `ImmutableSearchResponse`.

I also implemented logic similar to `addToHistory` for quickly loading novels that were already in the bookmarks.

Lastly, I added a very basic skeleton for these quickly loaded novels because, since they still depend on `ResultCached` instead of `ImmutableSearchResponse`, there are some parameters that don't exist in `ResultCached`. I think this could be improved.

TJMolina:
I fixed some small bugs, modified `DownloadViewModel2` and the history to send `ImmutableSearchResponse` directly so it doesn't need to be converted, and improved the `ResultScreen` skeleton.

LagradOst:
This is a way too large pr to give an accurate review. However here are some general comments

1. Never ever store a resource like statusRes, as resource ids are not stable between builds.
2. Always use PersistentList/PersistentMap when doing compose. List fucks up the compose compiler like `List<DefaultBookmark>` because compose does not know if it was changed between compositions
3. Do not store data like "deleteTarget: ImmutableSearchResponse" as it is more flexable to send the specific ImmutableSearchResponse to the action deleteNovel
4. Everything user facing should be immutable and marked as `@Immutable` (besides enums as they are already immutable) This ensures that the viewmodel does not have threading issues, signals that other devs should not add a var, and the compose compiler can use it for some magic.
5. cleanLegacyCorruptBookmarks? Why?

TJMolina:
about point 5 i forgot to delete that, it's from old version

LagradOst:
While I appreciate the PR, I feel like it is a bit too early to touch ResultViewModel2. I think you understood the idea I was going for with partial loading, but I am not a huge fan of the implementation. I would much rather finalize the ResultViewModel2 + View with support for partial loading myself before accepting a huge change to the logic. 

Ideally this pr should *only* be a +-500 line change that only changes minor UI stuff.

TJMolina:
sure no problem

PR #532, opened 2026-08-25
Refactor translation system

TJMolina:
This is a partial implementation of what I'm developing to improve the translator. In this case, it's meant to fix the issue where it's not working right now.

One important change is that it now uses the raw HTML to try to preserve the styles of the tags, and then wraps them in spans.


TJMolina:
Apparently, even these changes don't fix the issue. Google seems to have added some weird rate limit or an unusual anti-bot system. I need to investigate further to fix it. Also, this PR could be smaller by separating the functionalities, so I'll redo it.

PR #542, opened 2026-08-30
Refactor translation system

TJMolina:
Now this PR actually fixes the issue with the translator. It seems Google updated its anti-bot system, and it now detects when you make a lot of requests and then suddenly stop making them, temporarily banning you. The weird thing is that if you keep making requests continuously without stopping, you don't get blocked.

Obviously, that would be a pretty bad way to fix the issue, so after a lot of testing, I came up with the following approach: rebuild the client every time a request fails. Originally, I tried just rotating user agents, using WebView, cookies, etc., but nothing worked. This was the only thing that worked.

This PR is very, VERY stripped down to make it easy to test, but if you accept these changes, there are many more improvements coming afterward, such as support for tags in online translation and fast loading, so you won't have to wait 3 minutes per chapter when translating.


TJMolina:
The solution could be made more reliable by loading 50 chapters at a time, so you would have to rebuild the client fewer times. But with the current loading system, that's obviously a big NO. I already have an idea in mind for how to improve that, though.

LagradOst (approved):
Looks quite good, I might have some further comments when testing and rereviewing it but I think this is quite mergable!

